PR #101, opened 2014-09-16
replaced "console.log" with an optional dependence on "debug."

davidburhans:
I replaced the direct calls to "console" with a wrapper around "debug" in lib/debug.js. If "debug" is not found, it will just eat log messages.

If you would rather, I can have the lack of a dependence on "debug" to revert to use "console.log" instead of an empty function.


haio:
@crzidea  I like `debug` too.

PR #136, opened 2015-01-09
console.logs in highLevelConsumer

CharlesWall:
I notices some log statements in highLevelConsumer. These are spewing output when rebalancing and hindering adoption at my company. I've removed them in this commit.

### DIFF
--- a/lib/highLevelConsumer.js
+++ b/lib/highLevelConsumer.js
@@ -193,14 +193,12 @@ HighLevelConsumer.prototype.connect = function () {
     });
 
     function register() {
-        console.log("Registered listeners");
         // Register for re-balances (broker or consumer changes)
         self.client.zk.on('consumersChanged', rebalance);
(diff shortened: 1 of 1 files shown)


haio:
It's reasonable , but the log is useful anyway. I prefer use `debug` to reduce it.I think it will be more clean in next version, be calm

CharlesWall:
Perhaps we should allow the user to specify the log function they'd like to use with kafka and default to console.debug?

haio:
I don't think so, the logs in the module is not the users' application log, in most time they don't want it, as what you have mentioned above.

haio:
Now we can get the log with `DEBUG=kafka-node:* node sth`.

CharlesWall:
Oh now I see what you meant. This is a nice solution. Props.

haio:
Yep.

PR #140, opened 2015-01-15
Change to send the errors as the first parameter.

andersjanmyr:
Having to get the second parameter for an error is really confusing.


haio:
Yes, to not be zalgo, there are also some changes to be done in consumer and zk

andersjanmyr:
Yes, sorry about that. I updated the callers to use expect the same error patterns.

haio:
Thanks!

Issue #27, opened 2014-03-25
Working on supporting compression


haio:

jkrems:
Hey - I have a mostly working version of compression hacked into a fork right now. If there's interest, I would clean it up and open a PR.

jplock:
@jkrems we'd love to use it if you contribute it!

jkrems:
Created #141 for this.

haio:
The compression feature is available in ^0.2.16, thanks for @jkrems 's contribution.

PR #141, opened 2015-01-15
Handle snappy & gzip compression

jkrems:
Support for handling compression in both consumer and producer. On the consumer side it's handled transparently, for producers there's a new property in the payload (`attributes`) that can be set to 0, 1, or 2 (defaults to 0).

This PR also adds support for consumer.encoding which can be used to control how values are returned. Previously it was only possible to retrieve a stringified version of the value which destroyed binary encodings (like msgpack). The default is "utf8" which should match the previous behavior.

Fixes https://github.com/SOHU-Co/kafka-node/issues/27


haio:
@jkrems Nice implement! Need some time to test it.

PR #156, opened 2015-02-07
improve callback queue

estliberitas:
The addition solves the problem of never-called callback function passed to, for example, `Producer#send`.

Before there was no connection between socket (connection to broker) and callbacks passed to methods which interact with Broker and expect response. So once Broker died (OOM, stopped by supervisor, whatever), client started infinite reconnect attempts.

The point is callback should be called once we get socket error like `ECONNRESET` / `ECONNREFUSED`. This way module user can handle error in place where he calls some method (instead of handling client 'error' or 'close' events).


haio:
I will look into this

PR #158, opened 2015-02-11
More improments for callabck queue

haio:
We should ignore it when get null from queue, this may happen when a broker is down and it's callback queue is cleared, but there is some buffer still parsing. related issue #156


estliberitas:
LGTM :+1:

PR #159, opened 2015-02-11
Prevent fetch if consumers become ready while closing

qix:
I was experiencing connections left open after calling `.close()`


haio:
Hi, hat version your are using, can you put some example to reproduce this, I didn't see the problem.

qix:
Not a great example but this is a hacked version of your HighLevelConsumer example that has the problem.

We saw this in production (particularly in tests) where shortly after a consumer was ready we would want to shut it down, and would get a `BrokerNotReady` exception.

    var kafka = require('../kafka');
    var HighLevelConsumer = kafka.HighLevelConsumer;
    var Client = kafka.Client;
    var argv = require('optimist').argv;
    var topic = argv.topic || 'topic1';
    var client = new Client('localhost:2181');
    var topics = [ { topic: topic }];
    var options = { autoCommit: true, fromBeginning: false, fetchMaxWaitMs: 1000, fetchMaxBytes: 1024*1024 };
    var consumer = new HighLevelConsumer(client, topics, options);

    consumer.on('ready', function() {
      setTimeout(function() {
        consumer.close();
      }, 0);
    });

qix:
I couldn't replicate the successful connection again with a short script, but this throws a 'BrokerNotAvailable' where after my patch there is no error:

    'use strict';

    var kafka = require('../kafka');
    var HighLevelConsumer = kafka.HighLevelConsumer;
    var Client = kafka.Client;
    var argv = require('optimist').argv;
    var topic = argv.topic || 'topic1';
    var client = new Client('localhost:2181/kafka');
    var topics = [ { topic: topic }];
    var options = { autoCommit: true, fromBeginning: false, fetchMaxWaitMs: 1000, fetchMaxBytes: 1024*1024 };
    var consumer = new HighLevelConsumer(client, topics, options);

    consumer.on('ready', function() {
      consumer.close();
    });
    return;

haio:
consumer should always listening error event

PR #163, opened 2015-02-17
Add a dockerized kafka for running tests

JustinTulloss:
It was not obvious how to run tests before this. This patch adds a
simple docker setup so you can run tests without manually setting up
kafka.

Currently some of the zookeeper tests fail, but putting the pull request forward to see if the maintainers would be amenable to this change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,11 @@
 ALL_TESTS = $(shell find test -name 'test.*.js')
+KAFKA_TEST_HOST?=127.0.0.1
+
+kafka-instance:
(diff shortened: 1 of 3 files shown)


haio:
That sounds good, I will look into it. thanks!

LinusU:
`fig` is now renamed to `docker-compose`. `fig.yml` should be renamed to `docker-comopse.yml` and the command `fig` should be changed to `docker-compose`.

ismriv:
+1

hyperlink:
See my #387.

PR #164, opened 2015-02-17
Support sending a Message struct from the producer

JustinTulloss:
Currently the API does not offer a way to set the key on an outgoing
message. This resolves that hole by allowing devs to pass in Message
structs directly and have them be sent to the broker.

Has a test that passes.


haio:
Use KeyedMessage ca reduce user confusion.

JustinTulloss:
Looks good thanks!

PR #167, opened 2015-03-04
Allow null message values

JustinTulloss:
Message values may be null. If we get a null, pass -1 for the length.


haio:
Nice fix, thanks!

Issue #170, opened 2015-03-05
In case of `offsetOutOfRange`, the consumer should be paused


kadishmal:
The following is the output from the latest `"kafka-node": "0.2.21"`.

In our CI environment we test our code which has to send X messages to Kafka. Once all the messages are sent, we verify that Kafka has all the messages by reading all of them off of it. We assert that the X messages that we expect equal to the amount of messages read from Kafka. Pretty simple.

Now I've added some `console.log` to `kafka-node` `HighLevelConsumer` to show that when `offsetOutOfRange` event is emitted, we need to `pause` the consumer so that while we are adjusting the offset, the consumer doesn't try to fetch message because that emits yet another `offsetOutOfRange` event. The second one is unnecessary.

    $ node kafka.js
    Committed offset for "users-updated-db" topic is 179.
    The earliest message available in Kafka starts at 180 offset.
    The latest offset in Kafka is 188 meaning there are 8 messages in the queue
    8 need to be consumed.
    Going to consume messages from 180 offset for "users-updated-db" topic considering the earliest message available in Kafka.
    Going to fetch. Logged in `.fetch()`.
    offsetOutOfRange for topic 'users-updated-db'. Going to request the earliest offset available.
    'done` event is triggered after every fetch response. Received 0 messages.
    Going to fetch. `.fetch()` is called again.
    The earliest offset available in Kafka is 180. Going to set this offset to a topic.
    offsetOutOfRange for topic 'users-updated-db'. Going to request the earliest offset available.
    'done` event is triggered after every fetch response. Received 0 messages.
    Going to fetch. `.fetch()` is called again.
    The earliest offset available in Kafka is 180. Going to set this offset to a topic.
    'done` event is triggered after every fetch response. Received 0 messages.
    Going to fetch. `.fetch()` is called again.
    Message #1.
    'done` event is triggered after every fetch response. Received 1 messages.
    Going to fetch.
    Message #2. 
    Message #3. 
    Message #4. 
    Message #5. 
    Message #6. 
    Message #7. 
    Message #8. 
    'done` event is triggered after every fetch response. Received 8 messages.

As you can see `offsetOutOfRange` is triggered twice because `fetch` kept being called even though the first `offsetOutOfRange` event was emitted. This results in 2 requests to `self.offsetRequest([topic], function (err, offsets) {...`.

To avoid this it is necessary to pause the consumer as soon as `offsetOutOfRange` event is emitted.

[...]

PR #171, opened 2015-03-05
Fix #169: When paused why try to fetch every 1000 ms?

### LINKED ISSUE #169: When paused why try to fetch every 1000 ms?
In `highLevelConsumer.js` there is this code:

    this.on('done', function (topics) {
        self.updateOffsets(topics);
        if (!self.paused) {
            setImmediate(function () {
                self.fetch();
            });
        }
        else {
            setTimeout(function () {
                self.fetch();
            }, 1000);
        }
    });

The `.fetch()` function is as follows:

	HighLevelConsumer.prototype.fetch = function () {
	    if (!this.ready || this.rebalancing || this.paused) {
	        return;
	    }
	
	    this.client.sendFetchRequest(this, this.topicPayloads, this.options.fetchMaxWaitMs, this.options.fetchMinBytes, this.options.maxTickMessages);
	}

This means that even if the consumer is paused, it will try to fetch every 1 second:

           setTimeout(function () {
               self.fetch();
           }, 1000);

Why try if the code in `.fetch()` will just `return`?

            if (!this.ready || this.rebalancing || this.paused) {
	        return;
	    }

kadishmal:
Fixes #169 and #170.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,4 +2,5 @@
 
 ## Version NEXT (Unreleased)
 
+- Fix #170: In case of `offsetOutOfRange`, the consumer should be paused.
 - Fix #169: When paused why try to fetch every 1000 ms?
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,5 @@
+# kafka-node CHANGELOG
+
+## Version NEXT (Unreleased)
+
+- Fix #169: When paused why try to fetch every 1000 ms?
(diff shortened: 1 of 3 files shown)


haio:
The brace at the end of function doesn't need a semicolon, will clean it.

kadishmal:
The braces at the end of the function assignments are necessary because these functions are function expressions/assignments and not declarations.

haio:
Agree

PR #178, opened 2015-03-18
Initialize partition owner with consumerId

ismriv:
This avoids a race condition in `checkPartitionOwnership`, where another instance may create the node initialized to `null`, so the node will exist, it will try to get its contents but will fail in the comparison `consumerId !== data.toString()` since data may be `null`.

### DIFF
--- a/lib/zookeeper.js
+++ b/lib/zookeeper.js
@@ -326,7 +326,7 @@ Zookeeper.prototype.addPartitionOwnership = function (consumerId, groupId, topic
             function (callback) {
                 self.client.create(
                     path,
-                    null,
+                    new Buffer(consumerId),
                     null,
                     zookeeper.CreateMode.EPHEMERAL,
                     function (error, path) {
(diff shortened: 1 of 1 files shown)


haio:
Hi, what issue will this pr solve?

ismriv:
The following error:

```
TypeError: Cannot call method 'toString' of undefined
    at /activity-processor/node_modules/kafka-node/lib/zookeeper.js:395:53
    at /activity-processor/node_modules/kafka-node/node_modules/node-zookeeper-client/index.js:165:26
    ...
```

haio:
how to reproduce this error?

ismriv:
Having a few instances running, and restarting one of them to trigger the rebalancing mechanism. It's a race condition, so it may be tricky to reproduce, but if you try this a few times it will finally happen.

ismriv:
Any red flags?

haio:
Still can't reproduce it, but this a improvement logically, so I think it can be merged.

ismriv:
Brilliant! Thanks.

PR #190, opened 2015-04-01
Small performance improvement in topics diff

Mickael-van-der-Beek:
This pull requests slightly improves the performance of diff-ing the already created topics `topicMetadata`, with the list of topics that the user want's to create `topics`.

On average `a in b` should be faster than `b.indexOf(a)`.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -241,9 +241,8 @@ Client.prototype.createTopics = function (topics, isAsync, cb) {
         if (isAsync) return cb(null, 'All requests sent');
         var topicMetadata = resp[1].metadata;
         // ommit existed topics
-        var existed = Object.keys(topicMetadata);
         var topicsNotExists = topics.filter(function (topic) {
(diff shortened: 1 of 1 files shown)


haio:
Hi, the `in` operation will check inherited properties in a given object, I don't think it's faster than the `indexOf` of Array.

carlessistare:
I agree @haio indexOf should be faster, at least this test on my browser tells so:
https://jsperf.com/regex-vs-indexof-vs-in

estliberitas:
I think `in` op for object created with `Object.create(null)` will be faster than everything.

estliberitas:
... Because calls are expensive ops, even if they are optimized coz functions are in core of V8. And you do 2 `CallFunction` ops. here. While making an object w/ missing prototype would make `in` quite fast.

hyperlink:
Seems like a good change but could we include some benchmarks to verify?

PR #193, opened 2015-04-08
fix: emit the error instead of slient swallow it

xujihui1985:
I got kafka-node:zookeeper Exception: CONNECTION_LOSS[-4] error
sometimes, and the error sliently swallowed so it can not be catched

I thought the error should be thrown instead of been sliently swallowed.


haio:
Thanks! @xujihui1985

PR #196, opened 2015-04-14
Description: We want to return a 'broker not available' status when ther...

MaciekRakowski:
We want to emit an error when there is a problem with the socket connection to the kafka broker.


haio:
This is a nice fix, thanks!

PR #201, opened 2015-04-24
Exit cleanly from HighLevelConsumer

kjvalencik:
When using the `HighLevelConsumer`, the `ZooKeeper` client continues to recursively run `listConsumers` even after closing.

This causes an exception to be thrown if the process does not exit immediately.


kjvalencik:
The diff looks a little weird. It's better with whitespace ignored: https://github.com/SOHU-Co/kafka-node/pull/201/files?w=0

haio:
LGTM, I will test it.

PR #216, opened 2015-06-30
Do not update offsets when fetch request complete when autoCommit is set to false

anorsich:
In my project I use my own offset commit strategy and `autoCommit` is set to false. I maintain offset map in memory and update/commit offsets only when messages is processed, not when fetched.
Currently offsets are updated when every time when fetch request is completed. 

This PR checks if `autoCommit` is disabled and do not update offsets in this case. 

Thank you for your time.


haio:
If the offset isn't be updated, the client can't fetch more message from server.

PR #226, opened 2015-07-30
Passing Highwater mark down the chain to individual messages.

pcorliss:
Allows for fast-forwarding to current offset.


hyperlink:
This can be closed now since highwatermark attribute has been on the message for awhile now.

PR #231, opened 2015-08-07
Reduce connection timeout

dohse:


hpihkala:
+1

ismriv:
+1 here as well.

I'd pass the socket timeout setting in an options object instead, since it'll allow extend it for other reasons, such as https://github.com/SOHU-Co/kafka-node/pull/261

carlessistare:
Should consider writing some unit testing.
I had to deal with the same kind of issue in this PR #262 using fake timers of sinon.js

PR #242, opened 2015-09-03
fix(HighLevelProducer): allow sending to partition 0

simonvetter:
```js
p.partition = p.partition || self.client.nextPartition(p.topic);
```
will override p.partition with nextPartition() for a value of 0 because 0 is falsy.

This fixes it.


haio:
Thanks man!

PR #244, opened 2015-09-03
HighLevelProducer does not allow setting partition to 0

rgulewich:
If you send payloads with partition set to 0 using the HighLevelProducer, the partition number is ignored, and it just uses the next partition from `client.nextPartition()`.

This is the use case:
```javascript
var client = new kafka.Client(/* ... */);
var producer = new kafka.HighLevelProducer(client);



// Once clients are connected...



var payloads = [
    {
        topic: 'test',
        messages: [ 'message one' ],
        partition: 0
    },
    {
        topic: 'test',
        messages: [ 'message two' ],
        partition: 0
    }
];

producer.send(payloads[0], function (err, res) {
    // Gets sent to partition 0

    producer.send(payloads[1], function (err, res) {
        // Gets sent to partition 1

    });
});
```

### DIFF
--- a/lib/highLevelProducer.js
+++ b/lib/highLevelProducer.js
@@ -83,7 +83,7 @@ HighLevelProducer.prototype.send = function (payloads, cb) {
 HighLevelProducer.prototype.buildPayloads = function (payloads) {
     var self = this;
     return payloads.map(function (p) {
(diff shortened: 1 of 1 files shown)


haio:
Dup with #242

PR #265, opened 2015-10-29
encode -1 when key is empty or is null

cooper6581:
Allow for null keys.  We ran into issues running mirrormaker where the keyedmessages on the source broker had blank strings for the key instead of null.  This caused all the messages to be hashed to the same partition when mirrormaker republished.


RenEvo:
+1 on this.

PR #282, opened 2015-12-04
Replace node-uuid with a pure-JS implementation

d00rman:
The Apache Cassandra [driver](https://github.com/datastax/nodejs-driver) contains pure-JS UUID v1/v4 generators, which have been extracted into a separate repository - [`cassandra-uuid`](https://github.com/Pchelolo/uuid). Since the generation is used only in the HighLevelConsumer class, replace the binary module dependency with a pure one.

### DIFF
--- a/lib/highLevelConsumer.js
+++ b/lib/highLevelConsumer.js
@@ -5,7 +5,7 @@ var util = require('util'),
     events = require('events'),
     Client = require('./client'),
     protocol = require('./protocol'),
-    uuid = require('node-uuid'),
+    uuid = require('cassandra-uuid').Uuid,
     Offset = require('./offset'),
     async = require("async"),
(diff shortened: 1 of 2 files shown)


estliberitas:
@d00rman typo in `require(...)`

d00rman:
Duh, thank you @estliberitas ! Fixed.

hyperlink:
Looks like node-uuid is written in pure JS...

ctavan:
I have authored the implementation of v1 uuid's in the `node-uuid` module which has recently been renamed into `uuid` (see https://www.npmjs.com/package/uuid ) and I can confirm that it is a pure javascript implementation which performs really well and is extremely well tested. I suggest to simply use `uuid`.

PR #285, opened 2015-12-15
add in the ability to supply our own id for the consumer

wenisman:
Just a small change, this gives the ability to use a predefined consumer id if you want. 
Makes it easier to make a restful proxy, like if you are replacing the confluentinc one :)


ducas:
@wenisman you might want to resubmit from a cleaner branch...

PR #292, opened 2015-12-30
Add support for autoOffsetReset in high level consumer and default to largest

ducas:
Kafka's default offset for a consumer is the largest for the partition (https://kafka.apache.org/08/configuration.html). This change ensures that any new consumer with no initial offset in Zookeeper will reset to the largest offset by default.

### DIFF
--- a/lib/highLevelConsumer.js
+++ b/lib/highLevelConsumer.js
@@ -14,6 +14,8 @@ var util = require('util'),
     retry = require('retry'),
     debug = require('debug')('kafka-node:HighLevelConsumer');
 
+var AUTOOFFSETRESET_LARGEST = 'largest';
+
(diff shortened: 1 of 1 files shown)


Foo42:
Be good if there was also an option for time = -2, ie earliest available

sothychan:
Excellent.  I really need this feature.  Any idea when this would be merged?  Thanks!

hyperlink:
@sothychan I'm not planning on adding any new features to the `HighLevelConsumer`. You could  use the new [ConsumerGroup](https://github.com/SOHU-Co/kafka-node#consumergroup) (kafka 0.9+) which  has a `fromOffset` setting. Currently it only supports the latest offset but I'm in the process of updating it to support earliest as well for the next release.

PR #294, opened 2015-12-31
fromOffset is not used correctly on init

Hylke1982:
If you set the fromOffset to true the messages are not loaded from the offset, if set to false the message are loaded from the offset. If I read the documentation it should work the other way around.

### DIFF
--- a/lib/consumer.js
+++ b/lib/consumer.js
@@ -112,7 +112,7 @@ Consumer.prototype.init = function () {
             return self.emit('error', err);
         }
 
-        if (self.options.fromOffset)
+        if (!self.options.fromOffset)
             return self.fetch();
 
         self.fetchOffset(self.payloads, function (err, topics) {
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ node_modules
 *.swo
 test/data.txt
 docs
+
+*iml
+.idea/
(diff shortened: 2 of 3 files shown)


lib/consumer.js
```diff
@@ -112,7 +112,7 @@ Consumer.prototype.init = function () {
             return self.emit('error', err);
         }
 
-        if (self.options.fromOffset)
+        if (!self.options.fromOffset)
```
OndroNR: I think this should not be negated. If `fromOffset` is true, then provided offsets should be used. When you negate this condition, block will continue and fetch offsets from server.

Hylke1982: @OndroNR If you look at the rest of the (old) code when the fromOffsets is true, then the messages are fetched from kafka with offset = 0 and the function is exited with the return statement `return self.fetch();`

OndroNR: That function should be exited with `return self.fetch();` so it will not fetch commited offset with `self.fetchOffset`. Does it work for you with your fix? I can't force offset in payload regardless of `fromOffset` value. I have not yet tried your fix.

Hylke1982: @OndroNR I think I misinterpeted the fromOffset variable. (I assumed that setting the fromOffset to true that the offset for this client should be fetched from Kafka) Thank you for your feedback, I will close the pull request/

OndroNR: You are welcome. I still think cloning function argument in `Offset.prototype.buildPayloads` is good idea.

PR #295, opened 2016-01-03
add in the ability to supply our own id for the consumer

wenisman:
for multiple consumers it can be handy to be able to specify our own meaningful name for monitoring purposes.


haio:
This seems useful!

PR #301, opened 2016-01-08
Use synchronous decompression

### LINKED ISSUE #298: Wrong message order
I have problem with message order when consuming topic with 250k messages (which where loaded into Kafka in batches of 2000 messages) with fresh start (no offsets in zookeeper). Consumer often does not process messages from offset 0, rather it starts at 4000 or 8000, or so. Also it continuosly processes block of 1000 messages and jumps at later or sooner N*1000 offset. I have tried changed maxTickMessages to 800 and it process block od 800 messages, but it still jumped to N*1000 offset. I could not find missing 200 offsets in debug log. Chaning maxTickMessages or maxNumSegments to very large number did not help.

I was printing current message offset directly in Kafka binary protocol decoder, which should eliminate some of potential async effects. I am attaching [Offset log](https://github.com/SOHU-Co/kafka-node/files/81878/offsets2.tmp.txt) and used code [kafka-order-test.js](https://github.com/SOHU-Co/kafka-node/files/81875/kafka-order-test.js.txt). I think there is problem in Kafka binary protocol parsing, but I was not able to find problem in it.

Kafka itself should not be a problem as I dumped topic with kafkacat, which mantained correct offset and messages order. I also monitored node.js-Kafka network traffic with Wireshark, and messages were shown in correct order.

OndroNR:
This fixes #298 which was caused by asynchronous nested message set decompression which resulted in out of order message consuming. Kafka returns messages in MessageSet which contains nested compressed MessageSets of 2000 messages (in my testing). Unfortunately decompression was asynchronous without any synchronisation, so messages were processed out of order in batches of max 2000 (depends on maxTickMessages). My fix applies synchronous decompression.

### DIFF
--- a/lib/codec/index.js
+++ b/lib/codec/index.js
@@ -5,7 +5,8 @@ var zlib = require('zlib'),
 
 var gzipCodec = {
     encode: zlib.gzip,
-    decode: zlib.gunzip
+    decode: zlib.gunzip,
+    decodeSync: zlib.gunzipSync
 };
 
 var codecs = [
--- a/lib/codec/snappy.js
+++ b/lib/codec/snappy.js
@@ -37,5 +37,31 @@ function decodeSnappy(buffer, cb) {
     return snappy.uncompress(buffer, cb);
 }
 
+function decodeSnappySync(buffer) {
+    if (isChunked(buffer)) {
+        var pos = 16,
+            max = buffer.length,
+            encoded = [],
+            size;
+
+        while (pos < max) {
+            size = buffer.readUInt32BE(pos);
+            pos += 4;
+            encoded.push(buffer.slice(pos, pos + size));
+            pos += size;
+        }
+
+        var decodedChunks = encoded.map(function(data) {
+            var ret = snappy.uncompressSync(data);
+            return ret;
+        });
+
+        return Buffer.concat(decodedChunks);
+    }
+    return snappy.uncompressSync(buffer);
+}
+
 exports.encode = snappy.compress;
 exports.decode = decodeSnappy;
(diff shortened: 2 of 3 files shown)


estliberitas:
@OndroNR To say, replacing async with async-like sync is no good in general. This is more about how kafka-node does decompression, is it queued or not. And I guess, if you have such a problem, then it does something wrong - w/o queueing.

estliberitas:
More interested how this is implemented in Java version.

OndroNR:
@estliberitas would queued async decompression, which adds complexity, come with benefit here? Async decompression in kafka-node is missing queueing, resulting in loosing message order guarantee. Decompression is not I/O dependent, so I don't see any benefit in async.

estliberitas:
@OndroNR The benefit is not-blocking JS thread. I was trying to tell, that this is not the best solution, just trashing this library with simple workaround which should be solved in a more clean way.

OndroNR:
@estliberitas that make sense. My use case does not suffer from this problem. I will keep this PR here for reference.

aikar:
This PR can be closed. I've identified the bug that causes messages to be out of order, and it's simply a problem with the decoders logic. Asynchronous can work and still come in order.

I'm going to try to PR a fix next week.

OndroNR:
This bug is still present in library after such time? We are not consuming Kafka in node.js anymore. I just kept this PR open for reference. I will close it when you create PR.

aikar:
@OndroNR might as well go ahead and close it since this PR isn't even mergeable anymore.

OndroNR:
@aikar Alright, closing

PR #302, opened 2016-01-08
Increase default maxTickMessages

OndroNR:
Kafka returns messages in MessageSet which contains nested compressed MessageSets of 2000 messages (in my testing). Default `maxTickMessages` of 1000 in HighLevelConsumer causes to drop one half of every nested MessageSet messages and also could drop nested MessageSets itself if there would be enough of them. Increasing `maxTickMessages` to 10000 should ensure messages are not dropped .

### DIFF
--- a/lib/highLevelConsumer.js
+++ b/lib/highLevelConsumer.js
@@ -26,7 +26,7 @@ var DEFAULTS = {
     maxNumSegments: 1000,
     fetchMinBytes: 1,
     fetchMaxBytes: 1024 * 1024,
-    maxTickMessages: 1000,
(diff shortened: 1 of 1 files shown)


OndroNR:
Looks like, Kafka chooses MessageSet size dynamically. Kafka used 252 messages per MessageSet in another testing with single byte messages.

haio:
So this change seems not useful

OndroNR:
@haio it is useful, but only as workaround. I don't think maxTickMessages mechanism is well implemented, which results in messages dropping.

PR #328, opened 2016-02-11
fix bug where different partitoners where affecting others

### LINKED ISSUE #327: Bug when using multiple high level producers
Hi,

When instantiating multiple high level producers in the same nodejs process, each of them is sending a message to a different partition.

It happens because the following function is the same for all producers, meaning that they each modify the same `c` variable...

```javascript
CyclicPartitioner.prototype.getPartition = (function cycle() {
    var c = 0;
    return function (partitions) {
      if (_.isEmpty(partitions)) return 0;
      return partitions[ c++ % partitions.length ];
    };
})();
```

shaharmor:
fixes #327

### DIFF
--- a/lib/partitioner.js
+++ b/lib/partitioner.js
@@ -19,16 +19,14 @@ DefaultPartitioner.prototype.getPartition = function (partitions) {
 };
 
 var CyclicPartitioner = function () {
+    this.c = 0;
 };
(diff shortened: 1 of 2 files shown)


shaharmor:
@haio - any chance to get this merged?

PR #333, opened 2016-02-18
Fixed documentation and made autoCommit() force parameter optional.

cedarm:
I'm guessing that force is supposed to be optional.  I'll revise if not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,15 +302,16 @@ consumer.removeTopics(['t1', 't2'], function (err, removed) {
 });
 ```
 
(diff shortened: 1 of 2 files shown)


haio:
@cedarm could you also fix the problem in low level Consumer, thanks!

haio:
#306 has handled this, so close.

mllanes:
I can't seem to find this PR reflected on the docs

cedarm:
#306 was merged instead of this

Issue #339, opened 2016-03-15
for topic of multiple partitions, sometimes not all partitions are consumed by highlevelconsumer


tomwang1013:
it happened quite a few times: i use highlevelconsumer to consume a topic of 4 partitions, but only 2 partitions were consumed, like:
![image](https://cloud.githubusercontent.com/assets/1826003/13776506/7a8834c0-eae5-11e5-97e7-3130cc9f404d.png)

Sometime reboot the consumer fixed it, sometimes not. why?

hyperlink:
Does this issue happen in version **0.2.30** ?

tomwang1013:
@hyperlink： yes

sensoroDj:
I also encountered this situation。but in version 0.2.29 no this problem

sensoroDj:
@tomwang1013  and what tools do you use to see the partitions? thank you

tomwang1013:
[](https://github.com/yahoo/kafka-manager)

tomwang1013:
@sensoroDj  are you sure that 0.2.29 not has this problem? i will revert to 0.2.29

sensoroDj:
Sorry . I use the 0.2.29 version!  my partitions and consumers are one-for-one !  I will let one consumer consume one partitions eg: 3 patitions to 3 consumers  !   but when i use 0.3.x highlevel api  will not work i want to be. I recommand you to use low version when you use highleve api. I found it cant work well in 0.3.x

jbarreto:
@tomwang1013  can you give the output using kafka tools ?

Something like 
`
/opt/kafka/kafka_2.11-0.8.2.1/bin/kafka-run-class.sh kafka.tools.ConsumerOffsetChecker --group kafka-node-group --zookeeper [ZKHOST:ZKPORT] --topic [TOPIC_NAME]
`
example:
`
/opt/kafka/kafka_2.11-0.8.2.1/bin/kafka-run-class.sh kafka.tools.ConsumerOffsetChecker --group kafka-node-group --zookeeper local.zookeeper.net:2181 --topic feedUpdates2
`

The only scenario I have been able to reproduce your issue is when you are only sending messages to a few partitions and you don'f have one consumer per partition, for example 

- Publishing  to partitions 0,1,2,3,4
- Running 3 consumers 

You will get something like:
`
Group           Topic                          Pid Offset          logSize         Lag             Owner
[...]

PR #352, opened 2016-04-14
Fix fetchMaxBytes in example to match the default

hyperlink:
This should help with issues like #339 when people copy and paste from the examples.


estliberitas:
LGTM

PR #360, opened 2016-04-29
Fixes protocol value checking before converting to string

### LINKED ISSUE #359: Error while producing messages
It seems the value is undefined and is not being catched by the code before it, which checks for null only.

```js
if (value !== null) { // <===== here
        if (Buffer.isBuffer(value)) {
            m.Int32BE(value.length);
        } else {
            if (typeof value !== 'string') value = value.toString();
            m.Int32BE(Buffer.byteLength(value));
        }
        m.string(value);
    } else {
        m.Int32BE(-1);
    }
```
This results on this error when I run the example provided on the readme.
```
/Users/cantidio/Projects/node-kafka-rx/node_modules/kafka-node/lib/protocol/protocol.js:385
            if (typeof value !== 'string') value = value.toString();
                                                        ^

TypeError: Cannot read property 'toString' of undefined
    at encodeMessage (/Users/cantidio/Projects/node-kafka-rx/node_modules/kafka-node/lib/protocol/protocol.js:385:57)
    at /Users/cantidio/Projects/node-kafka-rx/node_modules/kafka-node/lib/protocol/protocol.js:358:19
    at Array.forEach (native)
    at encodeMessageSet (/Users/cantidio/Projects/node-kafka-rx/node_modules/kafka-node/lib/protocol/protocol.js:357:16)
    at /Users/cantidio/Projects/node-kafka-rx/node_modules/kafka-node/lib/protocol/protocol.js:345:30
    at Array.forEach (native)
    at /Users/cantidio/Projects/node-kafka-rx/node_modules/kafka-node/lib/protocol/protocol.js:344:14
    at Array.forEach (native)
    at _encodeProduceRequest (/Users/cantidio/Projects/node-kafka-rx/node_modules/kafka-node/lib/protocol/protocol.js:338:12)
    at /Users/cantidio/Projects/node-kafka-rx/node_modules/kafka-node/lib/protocol/protocol.js:326:16
```

When I fixed this check, everything works as expected.

cantidio:
fixes #359

### DIFF
--- a/lib/protocol/protocol.js
+++ b/lib/protocol/protocol.js
@@ -378,7 +378,7 @@ function encodeMessage(message) {
 
     var value = message.value;
 
-    if (value !== null) {
+    if (value !== null && value !== undefined) {
         if (Buffer.isBuffer(value)) {
             m.Int32BE(value.length);
         } else {


lib/protocol/protocol.js
```diff
@@ -378,7 +378,7 @@ function encodeMessage(message) {
 
     var value = message.value;
 
-    if (value !== null) {
+    if (value !== null && value !== undefined) {
```
hyperlink: this can be written as:
```js
if (value != null) {
```

estliberitas:
LGTM except minor nit described by @hyperlink

cantidio:
@estliberitas @hyperlink I would agree with this, but sometimes this type of abstraction is not very good.
For instance all comparisons on the project are made using strict checking with '==='. Then we start introducing abstract checking it makes things confusing. 

For instance you can open the same file in the future for a refactor and think that this '==' is a mistake and change it back to '==='. Tests could help preventing this, but that's my only concern. Linters would complain about abstract checking as well.

It's more of a question of abstract  vs strict checking.

But up to you guys, I can change it if you really thing that my point doesn't make sense. :)

PR #365, opened 2016-05-09
added path prefix for zookeeper key paths

SirCameron:


lib/zookeeper.js
```diff
@@ -305,29 +311,29 @@ Zookeeper.prototype.deletePartitionOwnership = function (groupId, topic, partiti
...
 Zookeeper.prototype.addPartitionOwnership = function (consumerId, groupId, topic, partition, cb) {
-    var path = '/consumers/' + groupId + '/owners/' + topic + '/' + partition,
+    var path = this.pathPrefix+'/consumers/' + groupId + '/owners/' + topic + '/' + partition,
         self = this;
 
     async.series([
             function (callback) {
                 self.client.create(
-                        '/consumers/' + groupId,
+                        this.pathPrefix+'/consumers/' + groupId,
```
hyperlink: What is the `this` context here? Should this line, 328, and 336 be using `self` instead of `this`?

lib/zookeeper.js
```diff
@@ -16,6 +16,12 @@ var zookeeper = require('node-zookeeper-client'),
  */
 var Zookeeper = function (connectionString, options) {
     this.client = zookeeper.createClient(connectionString, options);
+    if (typeof options!='undefined' && typeof options.pathPrefix!='undefined') {
```
hyperlink: Looks the new `pathPrefix` option is passed onto zookeeper. How does zookeeper handle that option? Could we move this block up one line and delete the pathPrefix if it was defined?

lib/zookeeper.js
SirCameron: Correct. I fixed it for lines 320, 328 and 336.

lib/zookeeper.js
SirCameron: I corrected this.

lib/zookeeper.js
```diff
@@ -30,7 +36,7 @@ var Zookeeper = function (connectionString, options) {
 util.inherits(Zookeeper, EventEmiter);
 
 Zookeeper.prototype.registerConsumer = function (groupId, consumerId, payloads, cb) {
-    var path = '/consumers/' + groupId;
+    var path = this.pathPrefix+'/consumers/' + groupId;
```
hyperlink: A nit here but could we keep the spacing before and after the `+` operator consistent with the previous code? In the future we will need to add JSCS to help enforce this.

SirCameron: ok

carlessistare:
git rebase -i HEAD^^^^  for your 4 commits would be nice

hyperlink:
But @carlessistare we have squashed merge! :)

lib/zookeeper.js
```diff
@@ -15,6 +15,14 @@ var zookeeper = require('node-zookeeper-client'),
  * @constructor
  */
 var Zookeeper = function (connectionString, options) {
+    if (typeof options!='undefined' && typeof options.pathPrefix!='undefined') {
```
carlessistare: Some space around !=  plz

lib/zookeeper.js
```diff
@@ -15,6 +15,14 @@ var zookeeper = require('node-zookeeper-client'),
  * @constructor
  */
 var Zookeeper = function (connectionString, options) {
+    if (typeof options!='undefined' && typeof options.pathPrefix!='undefined') {
+        this.pathPrefix = options.pathPrefix;
+        delete options.pathPrefix;
+    }
+    else{
```
carlessistare: Another on in here

carlessistare:
@SirCameron @hyperlink  looking at documentation https://www.npmjs.com/package/node-zookeeper-client
I have the feeling that this prefix can be added in connection string. (Ex: localhost:2181,localhost:2182/prefixPath )
@SirCameron could you test (if you didnt do it yet) and give us a feedback?

hyperlink:
That's great @carlessistare it looks like chroot is a zookeeper feature. https://zookeeper.apache.org/doc/r3.2.2/zookeeperProgrammers.html#ch_zkSessions

Learned something new today.

SirCameronMcAlpine:
Wow, I learned something also. Ok will check this.

SirCameronMcAlpine:
Ok, well guys, that was fun :) I tested that and it is the truth. Thanks @carlessistare.
We can close this pull request.

Issue #342, opened 2016-03-28
How to set consumer option "auto.offset.reset" ?


nobusue:
I'd like to consume from latest message when I don't have committed offset (ex: new consumer group).
At v0.3.2, HighLevelConsumer consumes from oldest message, but it isn't expected.

haio:
You can set fromOffset true and give the offset you wish to consume in payload, example:
```
consumer = new HighLevelConsumer(
        client,
        [
            { topic: 't', offset: 1000 }
        ],
        { fromOffset: true }
    );
```

nobusue:
Thanks haio, but it doesn't match my needs.
I just intend to start consume latest message for "fresh consumer" (= there is no committed offset entry in zookeeper). In other language like Java, Kafka client API supports to set option map, like "auto.offset.reset=latest".
Is there any way to set Kafka client API option map directly?

haio:
This feature is not supported at the moment.

CharlesWall:
I'll have a PR coming for picking up from the latest offset.

shubsk:
@CharlesWall any option like auto.offset.reset=latest , sometime soon for 0.8.*  version of kafka for node client.

hyperlink:
@shubsk I wouldn't hold my breath waiting for that.

CharlesWall:
I haven't looked at any of this stuff since May. I learned how this module worked to write the fix I needed. @shubsk You should consider doing the same.

hyperlink:
The solution is to use `ConsumerGroup` which supports both `earliest` and `latest`. I don't think it's worth updating the HLC since it has been deprecated by Kafka project.

PR #372, opened 2016-05-16
add offset#getLatestOffsets function to get all the latest offsets fr…

CharlesWall:
…om a group of topics populating those topics partitions

### DIFF
--- a/lib/highLevelConsumer.js
+++ b/lib/highLevelConsumer.js
@@ -34,7 +34,7 @@ var nextId = (function () {
     var id = 0;
     return function () {
         return id++;
-    }
+    };
 })();
 
 var HighLevelConsumer = function (client, topics, options) {
@@ -54,6 +54,7 @@ var HighLevelConsumer = function (client, topics, options) {
     this.connect();
 };
 util.inherits(HighLevelConsumer, events.EventEmitter);
+
 
 HighLevelConsumer.prototype.buildPayloads = function (payloads) {
     var self = this;
@@ -94,7 +95,7 @@ HighLevelConsumer.prototype.connect = function () {
         // Check the topics exist and create a watcher on them
         var topics = self.payloads.map(function (p) {
             return p.topic;
-        })
+        });
 
         self.client.topicExists(topics, function (err) {
             if (err) {
@@ -121,8 +122,8 @@ HighLevelConsumer.prototype.connect = function () {
     this.checkPartitionOwnershipInterval = setInterval(function () {
         if (!self.rebalancing) {
             async.each(self.topicPayloads, function (tp, cbb) {
-                if (tp['partition'] !== undefined) {
-                    self.client.zk.checkPartitionOwnership(self.id, self.options.groupId, tp['topic'], tp['partition'], function (err) {
+                if (tp.partition !== undefined) {
+                    self.client.zk.checkPartitionOwnership(self.id, self.options.groupId, tp.topic, tp.partition, function (err) {
                         if (err) {
                             cbb(err);
                         }
@@ -226,7 +227,7 @@ HighLevelConsumer.prototype.connect = function () {
 
     this.client.on('reconnect', function(lastError){
         self.fetch();
-    })
+    });
 
     this.client.on('close', function () {
         debug('close');
@@ -273,7 +274,7 @@ HighLevelConsumer.prototype.rebalanceAttempt = function (oldTopicPayloads, cb) {
                 debug("HighLevelConsumer %s stopping data read during rebalance", self.id);
                 self.stop(function () {
                     callback();
-                })
+                });
             },
 
             // Assemble the data
@@ -294,10 +295,10 @@ HighLevelConsumer.prototype.rebalanceAttempt = function (oldTopicPayloads, cb) {
             function (callback) {
                 debug("HighLevelConsumer %s releasing current partitions during rebalance", self.id);
                 async.eachSeries(oldTopicPayloads, function (tp, cbb) {
-                    if (tp['partition'] !== undefined) {
+                    if (tp.partition !== undefined) {
                         async.series([
                                 function (delcbb) {
-                                    self.client.zk.checkPartitionOwnership(self.id, self.options.groupId, tp['topic'], tp['partition'], function (err) {
+                                    self.client.zk.checkPartitionOwnership(self.id, self.options.groupId, tp.topic, tp.partition, function (err) {
                                         if (err) {
                                             // Partition doesn't exist simply carry on
                                             cbb();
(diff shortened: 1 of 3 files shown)


CharlesWall:
@hyperlink here is the PR that we talked about

lib/offset.js
```diff
@@ -58,4 +59,32 @@ Offset.prototype.fetchCommits = function (groupId, payloads, cb) {
     }
     this.client.sendOffsetFetchRequest(groupId, this.buildPayloads(payloads),cb);
 };
+
+
+Offset.prototype.fetchLatestOffsets = function (topics, cb) {
```
CharlesWall: I'm open to a different name for this function.

lib/offset.js
```diff
@@ -58,4 +59,43 @@ Offset.prototype.fetchCommits = function (groupId, payloads, cb) {
...
+    var client = this.client;
+    async.waterfall([
+        function (callback) {
+            client.loadMetadataForTopics(topics, callback);
+        },
+        function (topicsMetaData, callback) {
+            var payloads = [];
+            topics = topicsMetaData[1].metadata;
```
hyperlink: var?

hyperlink:
Wonderful PR @CharlesWall!  Do you have any plans to build this functionality into the consumer (as requested in #342) ? It would be vastly more usable there. It would be nice to be specify as a new `fromOffset` option.

```javascript
{
 fromOffset: 'latest'
}
```

What do you think?

CharlesWall:
that is my ultimate goal.

lib/offset.js
CharlesWall: ya that should be added.

lib/highLevelConsumer.js
```diff
@@ -55,6 +55,7 @@ var HighLevelConsumer = function (client, topics, options) {
 };
 util.inherits(HighLevelConsumer, events.EventEmitter);
 
+
```
andrewaustin: extra whitespace?

hyperlink:
Any feedback on this PR @estliberitas ?

bstahlhood:
@CharlesWall Does this work as stated above:

{
 fromOffset: 'latest'
}

?

nitsnwits:
I have the same question as @bstahlhood . `fromOffset: 'latest'` is not working for me. I think developers are supposed to call the method, `fetchLatestOffset` [method here](https://github.com/SOHU-Co/kafka-node/blob/d7f3583767d004518fa00122cbf2692133924fbe/lib/offset.js#L64) and pass the latest offset value in the options `fromOffset` while spinning up a consumer. 
Any suggestions?

hyperlink:
Right @nitsnwits. You need to do this [manually](https://github.com/SOHU-Co/kafka-node#for-a-new-consumer-how-do-i-start-consuming-from-the-latest-message-in-a-partition) until someone submits a PR for this feature.

nitsnwits:
Thanks @hyperlink 
This is fine if there is only one partition for the topic. If there are multiple partitions, and I want to start the consumer on the latest offset, then it becomes a problem. I will know which offset which partition is at from `fetchLatestOffsets`, however, I will not know from which partition the consumer will start consuming from because the ownership is gained after rebalancing. I am not sure if there is a way to provide `fromOffset` values for all partitions. Can you suggest me a way to do that? I want to spin up as many consumers as the number of partitions, which is the usual configuration for busy Kafka streams, and then tackle, if it's a new consumer, then start it from the latest offset. 

Secondly, to test if it's a new consumer group being registered, I am trying to `offset.fetch` and see if the current offset in every partition is -1, which indicates new consumer group. Please let me know if you have some suggestions on this.

Thanks again for help.

hyperlink:
@nitsnwits I would call `client.sendOffsetCommitRequest` and pass it the groupId, and latest offsets for each partition so that way when your consumers start up they can consume from those offsets. And In this case it would not matter who owns the partition since when your group starts it will consume from the latest offset that you set.

To see if a new consumer group is registered you could look in zookeeper and see if this znode `/consumers/<GroupID>` exists. A group that doesn't exist should not have that znode.

nitsnwits:
@hyperlink Thanks for the info. Sorry for a question so late on this. I am able to figure out if it's a new consumer group or not with the ZNode. That works out pretty well. 
I'm not sure in what format `client.sendOffsetCommitRequest` accepts it's payloads.

This is what I get from `offset.fetchLatestOffsets`.

```javascript
{
  "topic": {
    "0": 700
  }
}
```

I have tried different combinations of payloads to `sendOffsetCommitRequest`. 

```javascript
client.sendOffsetCommitRequest( [ { topic: 'topic', offsets: results_from_above } ], callback);
```

It always throws up an error. Can you suggest me how do I form the payload for this method. Either it's not able to find an array or gives `BrokerNotAvailable: could not find the leader error`.

I think this is a good topic for Troubleshooting/FAQ in ReadMe. I can send a Pull Request to update the ReadMe as a I figure it out.

Thanks for your help. Much appreciated.

hyperlink:
@nitsnwits I have steered you wrong when I mentioned the znode. I realize now future proof-ness will be an issue with that approach. Starting from 0.9 on kafka clients are not reliant on zookeeper and all requests are done by calling kafka directly.

I think the correct thing to do is to call your consumer's `fetchOffset` method or if you're not using a consumer call `client.sendOffsetFetchRequest`. These methods call to kafka for the information. The response will contain the offsets from the requested topic/partitions. If offsets have never been saved then the offset response will be `-1`.  Going forward this will be the best approach.

Also I am in the process of working on a GroupConsumer which takes advantage of the 0.9+ consumer coordinator and it does not interact with zookeeper. I hope to have a PR ready for it next week if you are interested in it.

Re: your code it looks like the first argument of the sendOffsetCommitRequest should be the `group_id`.

nitsnwits:
@hyperlink Thanks for the info. I'm still working with 0.8 so I've to work with it for now. :) 

Yes, I have the groupId as the first argument, sorry I missed to include that. But the payloads are still not accepted by the method `sendOffsetCommitRequest`. 
By using fetchOffset method, I am able to fetchLatestOffsets for all the partitions, but I am not able to save those offsets as the new consumer registers. 
My problem is that I have a new consumer registering to a very busy kafka stream, and I want it to start from the latest offsets instead of consuming from the start and bursting other upstream systems.
Is there a better way to do that? Thanks in advance!

hyperlink:
I think you are on the right track:

1. Get the current offsets for the group
2. if there are any offsets with -1 then call get latest offsets and update
3. commit the new offsets and start the consumer

The payload for `sendOffsetCommitRequest` should look like this:

```json
[
  {
    "topic": "RebalanceTopic",
    "partition": 1,
    "offset": 13,
    "maxBytes": 1048576,
    "metadata": "m"
  },
  {
    "topic": "RebalanceTest",
    "partition": 0,
    "offset": 5,
    "maxBytes": 1048576,
    "metadata": "m"
  }
]
```

There's a (undocumented) method called `buildPayloads` that will build the above structure for you given an array of topics.

PR #375, opened 2016-05-19
Consume large messages. Optimised solution for receiving socket packets.

subzero10:
Client.js -> socket.on('data') + Client.handleReceivedData
new Class -> SmartBuffer.js


crzidea:
Actually, I have a plan to remove all `Buffer#slice()`/`Buffer#concat()` usage in this module to improve performance. But it is still in reviewing. See #654

Issue #350, opened 2016-04-06
[kafka-node 0.3.2] When I restart kafka-node high-level consumer, consumer fetch messges with offset 0


baolinCloud:
Each  time  I restart kafka-node high-level consumer, I can see the error log from the server as belows:

ERROR [KafkaApi-0] Error when processing fetch request for partition [test,0] offset 0 from consumer with correlation id 66276 (kafka.server.KafkaApis)
kafka.common.OffsetOutOfRangeException: Request for offset 0 but we only have log segments in the range 45503 to 51197.
        at kafka.log.Log.read(Log.scala:380)
        at kafka.server.KafkaApis.kafka$server$KafkaApis$$readMessageSet(KafkaApis.scala:530)
        at kafka.server.KafkaApis$$anonfun$kafka$server$KafkaApis$$readMessageSets$1.apply(KafkaApis.scala:476)
        at kafka.server.KafkaApis$$anonfun$kafka$server$KafkaApis$$readMessageSets$1.apply(KafkaApis.scala:471)
        at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
        at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
        at scala.collection.immutable.Map$Map1.foreach(Map.scala:109)
        at scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
        at scala.collection.AbstractTraversable.map(Traversable.scala:105)
        at kafka.server.KafkaApis.kafka$server$KafkaApis$$readMessageSets(KafkaApis.scala:471)
        at kafka.server.KafkaApis$FetchRequestPurgatory.expire(KafkaApis.scala:783)
        at kafka.server.KafkaApis$FetchRequestPurgatory.expire(KafkaApis.scala:765)
        at kafka.server.RequestPurgatory$ExpiredRequestReaper.run(RequestPurgatory.scala:216)
        at java.lang.Thread.run(Thread.java:745)

hyperlink:
Should be fixed in version 0.5.1.

Issue #235, opened 2015-08-26
HighLevelConsumer can fetch messages from 0 offset if resume called before rebalancing complete


IharKrasnik:
Hello, in our project we experienced problem when one of our consumers reconsumed all topic messages from 0 offset. 
After digging in logs I've found that it happened immediately after one of rebalances (before that moment rebalances happened periodically and everything worked fine, offsets were ok).

I found that rebalanced event is emitting before topicPayloads set to real offsets (newTopicPayloads have 0 offset before fetchOffset completed) https://github.com/SOHU-Co/kafka-node/blob/master/lib/highLevelConsumer.js#L109

So, if pause consumer from outside and rebalancing happens and then consumer.resume() called before fetchOffset completed, fetch will be called with current topicPayloads that will have 0 offset for that moment and messages will be reconsumed
https://github.com/SOHU-Co/kafka-node/blob/master/lib/highLevelConsumer.js#L593
https://github.com/SOHU-Co/kafka-node/blob/master/lib/highLevelConsumer.js#L485

Am I missing something or is it a real issue?

jezzalaycock:
I have written a job manager which help manage offsets and commit only when the client has completed processing of a message - see https://www.npmjs.com/package/kafka-node-manager. 

Managing committing of offsets is a real pain and quite thorny.  Hopefully that job manager package will help you.

ersumitchh:
Having the same problem and it generally happens when lag size is zero.

ersumitchh:
Please fix this.

darthjoel:
Have exact same problem.  Trying jezzalaycock's job manager.

bencebalogh:
@hyperlink I think this can be closed as it should be fixed with #402

PR #376, opened 2016-05-20
Update highLevelConsumer.js

zoluzo:
If highLevelConsumer try to fetch a message when it is rebalancing, it maybe cause consumer reconsume message from min offset.
I find some issues:
https://github.com/SOHU-Co/kafka-node/issues/350
https://github.com/SOHU-Co/kafka-node/issues/235

### DIFF
--- a/lib/highLevelConsumer.js
+++ b/lib/highLevelConsumer.js
@@ -143,7 +143,9 @@ HighLevelConsumer.prototype.connect = function () {
 
         if (!self.rebalancing) {
             deregister();
-
+            
+            self.ready = false;
+            
             self.emit('rebalancing');
 
             self.rebalancing = true;


bencebalogh:
We face the same issue as you described, but we haven't found a way to reproduce it to test properly. Can you please describe the steps to reproduce?

zoluzo:
@bencebalogh Sorry, I can not reproduce it 100%.
When I restart my server, the highLevelConsumer may be reconsume(10% ~ 20% possibility).
After I change the code, I restart the server more than 20 times and not run into the problem.

estliberitas:
LGTM

estliberitas:
cc @hyperlink

bencebalogh:
I managed to reproduce the issue as well, and can confirm that this solution would fix it. Now that I could reproduce I did some digging.

Order when everything works as intended:
1. rebalancing started, consumer.rebalancing set to true
2. rebalancing finished, consumer.rebalancing set to false
3. offsets updated in consumer.rebalanced event listener
4. fetch called with correct offset

However due how the done event is emitted in can create the following incorrect order: 
1. rebalancing started, consumer.rebalancing set to true
2. rebalancing finished, consumer.rebalancing set to false
3. done event emitted
4. fetch called with incorrect offset (0)
5. offsets updated in consumer.rebalanced event listener
6. fetch called with correct offset

The fix works, but I don't think consumer's ready state should be changed to false, I'd propose to set the rebalancing state to false if there was an error during rebalance (inside the rebalance function) or if there wasn't any error then change it false once the offsets were updated in the rebalanced event listener. This basically the same thing, but we modify only the only state, which is strictly related to rebalancing.

zoluzo:
@bencebalogh  
I agree the wrong order that you mention above. Very explicitly.
I think consumer`s ready state should be false when it can not fetch message from kafka server.

bencebalogh:
This would be fixed with #402 (same issue)

hyperlink:
Closing as mentioned above this should be fixed by #402

PR #377, opened 2016-05-23
watch partition change event

springuper:
Assign `HighLevelConsumer` the ability to rebalance whenever Kafka partitions of some topic changed.


hyperlink:
See #430 which resolves conflicts.

springuper:
got it, thx.

Issue #334, opened 2016-03-01
TLS/SSL support


atkinsnashville:
Please add support for implementing 0.9 SSL functionality in broker connections.

jadedfire:
+1

shpros:
Is there a target release or date (month, quarter, year) for this issue?

kumar-b:
+1

arunpmathew:
+1

austinkelleher:
+1

hyperlink:
@atkinsnashville @jadedfire @shpros @kumar-b @arunpmathew @austinkelleher would you be willing to test @CharlesWall's PR and see if it works for you?

shpros:
@hyperlink : Thanks for the notification. I'll have my colleague @jmacias try it out.

kumar-b:
we are facing some issues in TLS enabled cluster. #427 
Anyone else tested with TLS enabled cluster?

PR #383, opened 2016-06-16
enable ssl client connections to brokers (#334)

CharlesWall:
Implement ssl socket for client/broker connections. only works with +0.9


lib/client.js
```diff
@@ -60,20 +64,47 @@ var Client = function (connectionString, clientId, zkOptions, noAckBatchOptions)
...
-                var broker = brokers[key];
-                var addr = broker.host + ':' + broker.port;
-                self.brokers[addr] = self.createBroker(broker.host, broker.port);
+            .forEach(function (key, index) {
+                var brokerProfile = brokers[key];
+                var addr = brokerProfile.host + ':' + brokerProfile.port;
+                self.brokerProfiles[addr] = brokerProfile;
+                if (self.ssl) {
+                    var sslAddress = brokerProfile.endpoints.find(function(endpoint) {
+                        return endpoint.startsWith('SSL://');
```
hyperlink: this won't work in `0.12.14` and below.

lib/client.js
```diff
@@ -60,20 +64,47 @@ var Client = function (connectionString, clientId, zkOptions, noAckBatchOptions)
...
+                    });
+                    if(!sslAddress) {
+                        var err = new Error('No SSL kafka endpoint found');
+                        self.emit('error', err);
+                    }
+                    sslAddress = splitOnFirst(sslAddress, '://')[1];
```
hyperlink: Couldn't you use the `url` module to do the parsing?

test/test.consumer.js
```diff
@@ -29,368 +30,396 @@ function offsetOutOfRange (topic, consumer) {
...
+    var suiteTimeout = testParameters.suiteTimeout || null;
+    var suiteName = testParameters.name;
+
+
+    function createClient() {
+        var clientId = 'kafka-node-client-' + uuid.v4();
+        return new Client(host, clientId, undefined, undefined, sslOptions);
+    }
+
+    describe(suiteName, function () {
```
CharlesWall: everything after this is just indenting to put it inside of the forEach above.

docker/certs/ca-key
```diff
@@ -0,0 +1,18 @@
+-----BEGIN RSA PRIVATE KEY-----
```
CharlesWall: it feels wrong to commit up certs and a private key, but they are only for development use.

docker-compose.yml
```diff
@@ -8,10 +8,19 @@ services:
...
     ports:
       - "9092:9092"
+      - "9093:9093"
     depends_on:
       - zookeeper
     environment:
       KAFKA_ADVERTISED_HOST_NAME: ${KAFKA_ADVERTISED_HOST_NAME}
       KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
+      KAFKA_LISTENERS: "PLAINTEXT://:9092,SSL://:9093"
+      KAFKA_ADVERTISED_LISTENERS: "PLAINTEXT://192.168.64.2:9092,SSL://192.168.64.2:9093"
```
hyperlink: you will need to change the hardcoded IP to `${KAFKA_ADVERTISED_HOST_NAME}`

lib/client.js
```diff
@@ -509,8 +535,23 @@ Client.prototype.setupBroker = function (host, port, longpolling, brokers) {
...
+
     var self = this;
-    var socket = net.createConnection(port, host);
+    var socket;
+    if (self.ssl) {
+        socket = tls.connect(port, host, self.sslOptions);
+    } else {
+        socket = net.createConnection(port, host);
+    }
     socket.addr = host + ':' + port;
```
hyperlink: Please use `addr` here so we don't have to concat `host` and `port` again.

README.md
```diff
@@ -16,11 +16,12 @@ Follow the [instructions](http://kafka.apache.org/documentation.html#quickstart)
 
 # API
 ## Client
-### Client(connectionString, clientId, [zkOptions], [noAckBatchOptions])
+### Client(connectionString, clientId, [zkOptions], [noAckBatchOptions], [sslOptions])
 * `connectionString`: Zookeeper connection string, default `localhost:2181/`
 * `clientId`: This is a user-supplied identifier for the client application, default `kafka-node-client`
 * `zkOptions`: **Object**, Zookeeper options, see [node-zookeeper-client](https://github.com/alexguan/node-zookeeper-client#client-createclientconnectionstring-options)
 * `noAckBatchOptions`: **Object**, when requireAcks is disabled on Producer side we can define the batch properties, 'noAckBatchSize' in bytes and 'noAckBatchAge' in milliseconds. The default value is `{ noAckBatchSize: null, noAckBatchAge: null }` and it acts as if there was no batch
+* `sslOptions`: **Object**, options to be passed to the tls broker sockets, ex. { rejectUnauthorized: false }
```
hyperlink: can we add a blurb about how this is only supported for kafka 0.9 and above?

lib/client.js
CharlesWall: `addr` may have changed because of :545/:546 it is important that we recalculate it here.

docker-compose.yml
CharlesWall: will fix.

lib/client.js
hyperlink: yup, I see that now. 👍

hyperlink:
Feedback on this PR @haio or @estliberitas ?

hyperlink:
Thank you for your contribution @CharlesWall !

PR #399, opened 2016-07-12
Fixes #398 where reconnect fails for SSL brokers

### LINKED ISSUE #398: Error: Cannot read property 'sslHost' of undefined (client.js:546:23)
We're using the new SSL support and saw this error when trying to create our Kafka producer:

```
Cannot read property 'sslHost' of undefined
    at connect [as createBroker] (/Users/philidem/Development/third-party/kafka-node/lib/client.js:546:23)
    at Client.setupBroker (/Users/philidem/Development/third-party/kafka-node/lib/client.js:535:24)
    at Client.reconnectBroker (/Users/philidem/Development/third-party/kafka-node/lib/client.js:610:24)
    at null._onTimeout (/Users/philidem/Development/third-party/kafka-node/lib/client.js:599:12)
    at Timer.listOnTimeout (timers.js:92:15)
```

I'm logging this issue here to keep track of it but I will try to add more details later.

Problem lines are these:

```javascript
host = brokerProfile[this.ssl ? 'sslHost' : 'host'] || host;
port = brokerProfile[this.ssl ? 'sslPort' : 'port'] || port;
```

I think this is the right fix:

```javascript
host = brokerProfile && brokerProfile[this.ssl ? 'sslHost' : 'host'] || host;
port = brokerProfile && brokerProfile[this.ssl ? 'sslPort' : 'port'] || port;
```

I don't know enough about _brokerFile_. I would like to understand a little better why it might be undefined in some cases before applying this fix. I don't want to hide another underlying problem.

hyperlink:


lib/client.js
```diff
@@ -120,6 +102,31 @@ Client.prototype.connect = function () {
...
+  Object.keys(brokers).forEach(function (key, index) {
+    var brokerProfile = brokers[key];
+    var addr = brokerProfile.host + ':' + brokerProfile.port;
+    self.brokerProfiles[addr] = brokerProfile;
+    if (self.ssl) {
+      var sslAddress = brokerProfile.endpoints.find(function (endpoint) {
+        return url.parse(endpoint).protocol === 'ssl:';
+      });
+      if (!sslAddress) {
+        var err = new Error('No SSL kafka endpoint found');
```
philidem: You probably should include more information about the specific broker that does not have an SSL endpoint. 

Also, does every broker need to have an SSL endpoint or do we need to simply find at least one broker that has an SSL endpoint? I'm worried that the `error` that you emit here will cause an exception (since `EventEmitter`, by default, will throw error if there is no listener for the `error` event) even if there are other valid brokers with SSL endpoint.

hyperlink: This method is a refactor from what @CharlesWall wrote. I have updated the error message to include the broker addr and id. I would expect all brokers in a cluster to be configured similarly to each other and it would be a big problem if one is missing an secure endpoint. Emitting an error is the right thing to do here.

PR #402, opened 2016-07-15
Prevent fetch before offset are updated after high level consumer rebalance

snatesan:
If there is done event fired on high level consumer 
after "rebalanced" event 
but before "updateOffsets" method in "rebalanced" event handler of HighLevelConsumer being fulfilled 
then that will trigger fetching all messages from zookeeper for the partition requested.
This is happening because at that stage owned partitions of HighLevelConsumer is set to 0.
This PR make sure "rebalancing" flag is set to "false" only after offsets are updated with the values from zookeeper and hence any "fetch" call before that will be ignored.

This issue can happen when rebalance occurs frequently in short time due to various reasons like network interruptions or addition or removal of a consumer with same group id.

### DIFF
--- a/test/test.highlevelConsumer.js
+++ b/test/test.highlevelConsumer.js
@@ -1,0 +1,195 @@
+'use strict';
+var util = require('util');
+var EventEmitter = require('events');
+var sinon = require('sinon');
+
+var libPath = '../lib/';
+
+var HighLevelConsumer = require(libPath + 'highLevelConsumer');
+
+function FakeZookeeper () {
+  EventEmitter.call(this);
+
+  this.checkPartitionOwnership = function (id, groupId, topic, partition, cb) {
+    setImmediate(cb);
+  };
+  this.deletePartitionOwnership = function (groupId, topic, partition, cb) {
+    setImmediate(cb);
+  };
+  this.addPartitionOwnership = function (id, groupId, topic, partition, cb) {
+    setImmediate(cb);
+  };
+  this.registerConsumer = function (groupId, id, payloads, cb) {
+    setImmediate(cb);
+  };
+  this.listConsumers = function (groupId) {};
+  this.getConsumersPerTopic = function (groupId, cb) {
+    var consumerTopicMap = {};
+    consumerTopicMap[groupId] = [ 'fake-topic' ];
+    var topicConsumerMap = { 'fake-topic': [ groupId ] };
+    var topicPartitionMap = { 'fake-topic': [ '0', '1', '2' ] };
+    var map = {
+      consumerTopicMap: consumerTopicMap,
+      topicConsumerMap: topicConsumerMap,
+      topicPartitionMap: topicPartitionMap
+    };
+
+    setImmediate(cb, null, map);
+  };
+}
+util.inherits(FakeZookeeper, EventEmitter);
+
+function FakeClient () {
+  EventEmitter.call(this);
+
+  this.zk = new FakeZookeeper();
+
+  this.topicExists = function (topics, cb) {
+    setImmediate(cb);
+  };
+  this.refreshMetadata = function (topicNames, cb) {
+    setImmediate(cb);
+  };
+  this.sendOffsetCommitRequest = function (groupId, commits, cb) {
+    setImmediate(cb);
+  };
+  this.sendFetchRequest = function (consumer, payloads, fetchMaxWaitMs, fetchMinBytes, maxTickMessages) {};
+  this.sendOffsetFetchRequest = function (groupId, payloads, cb) {
+    setImmediate(cb);
+  };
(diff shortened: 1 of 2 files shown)


hyperlink:
Thanks for the PR. Code looks good to me.

Can we add a unit test to verify `rebalancing` is false only after fetchOffset and updateOffsets are called?

snatesan:
@hyperlink Added unit test.

zoluzo:
@snatesan  Good work.
@hyperlink  Please close https://github.com/SOHU-Co/kafka-node/pull/376. I think this could fix the problem.

test/test.highlevelConsumer.js
```diff
@@ -0,0 +1,222 @@
...
+var libPath = process.env['KAFKA_COV'] ? '../lib-cov/' : '../lib/';
+
+// var Client = require(libPath + 'client');
+var HighLevelConsumer = require(libPath + 'highLevelConsumer');
+
+function FakeZookeeper () {
+  EventEmitter.call(this);
+
+  this.checkPartitionOwnership = function (id, groupId, topic, partition, cb) {
+    setTimeout(cb, 0);
```
hyperlink: All these `setTimeout` of 0 can be replaced with `setImmediate`

test/test.highlevelConsumer.js
```diff
@@ -0,0 +1,222 @@
+'use strict';
+var util = require('util');
+var EventEmitter = require('events');
+var sinon = require('sinon');
+
+var libPath = process.env['KAFKA_COV'] ? '../lib-cov/' : '../lib/';
+
+// var Client = require(libPath + 'client');
```
hyperlink: I think this was copied from another test. This comment should be deleted. And on the line below I am not sure what this `KAFKA_COV` was used for but we should delete that as well.

test/test.highlevelConsumer.js
```diff
@@ -0,0 +1,222 @@
...
+    it('should emit rebalanced event and clear rebalancing flag only after offsets are updated', function (done) {
+      client.emit('ready');
+
+      highLevelConsumer.on('registered', function () {
+        var sendOffsetFetchRequestCalled = false;
+        var refreshMetadataCalled = false;
+
+        // verify rebalancing is false until rebalance finishes
+        var refreshMetadata = client.refreshMetadata;
+        client.refreshMetadata = function () {
```
hyperlink: This can be shortened by using a stub. And later you can `sinon.assert` that this stub was called.

```js
        sandbox.stub(client, 'sendOffsetFetchRequest', function (groupId, payloads, cb) {
          highLevelConsumer.rebalancing.should.be.true;
          setImmediate(cb);
        });
```

test/test.highlevelConsumer.js
```diff
@@ -0,0 +1,222 @@
...
+          refreshMetadataCalled = true;
+          if (!highLevelConsumer.rebalancing) {
+            done(new Error('rebalancing is set to false before rebalance finishes'));
+            return;
+          }
+          refreshMetadata.apply(this, arguments);
+        };
+
+        client.sendOffsetFetchRequest = function (groupId, payloads, cb) {
```
hyperlink: Similarly here:

```js
        sandbox.stub(client, 'sendOffsetFetchRequest', function (groupId, payloads, cb) {
          highLevelConsumer.rebalancing.should.be.true;
          setImmediate(cb);
        });
```

test/test.highlevelConsumer.js
```diff
@@ -0,0 +1,222 @@
...
+            if (!highLevelConsumer.rebalancing) {
+              done(new Error('rebalancing flag is set to false before rebalance finishes'));
+              return;
+            }
+            cb();
+          }, 10);
+        };
+
+        highLevelConsumer.on('rebalanced', function () {
+          if (!sendOffsetFetchRequestCalled || !refreshMetadataCalled) {
```
hyperlink:
```js
          sinon.assert.calledOnce(client.refreshMetadata);
          sinon.assert.calledOnce(client.sendOffsetFetchRequest);
          sinon.assert.callOrder(client.refreshMetadata, client.sendOffsetFetchRequest);
          highLevelConsumer.rebalancing.should.be.false;
```

test/test.highlevelConsumer.js
snatesan: I tried to use Client initially. But it was a pain so eventually mocked it. Will delete the comment.
Regarding KAFKA_COV that is how libPath is being set in test.consumer.js. So I thought it was being set somewhere. Will remove it if it is not needed.

hyperlink: Yes please. I've been trying to clean up things in the code that don't make sense or isn't used anymore.

snatesan:
@hyperlink Thanks for the review. Addressed comments.

PR #422, opened 2016-08-01
Added code coverage and HighLevelProducer tests.

lightswitch05:
I added code coverage using istanbul. My tests were failing on Travis, after some research I found it was due to some changes with [NPMv3 and peer dependencies](https://travis-ci.org/lightswitch05/kafka-node/jobs/148754606#L296-L298) with `eslint-config-semistandard`. After getting that working, I saw `HighLevelProducer` was in sore need of some tests, so I added those. They are pretty much the same test as `Producer`. Testing my changes some more with Travis I saw that [proxyquire was not working](https://travis-ci.org/lightswitch05/kafka-node/jobs/148760235#L672-L678), I changed the module require to a file to fix it.

I would be happy to add integration with https://coveralls.io/ if there is any interest to have that level of code coverage information with pull requests. It would require the project owner to allow integration with the account.

Summary: Added code coverage. Added unit tests. [Fixed tests on travis](https://travis-ci.org/lightswitch05/kafka-node/builds/148936864).


test/test.client.js
```diff
@@ -18,7 +18,7 @@ describe('Client', function () {
     before(function () {
       zk = new FakeZookeeper();
 
-      Client = proxyquire('../lib/Client', {
+      Client = proxyquire('../lib/client.js', {
```
hyperlink: should be able to omit the `.js`

lightswitch05: You are right @hyperlink - I amended the commit and verified the build is still successful on Travis: https://travis-ci.org/lightswitch05/kafka-node/builds/148936864

hyperlink:
Thanks for the PR @lightswitch05!

I was thinking since one of the HLP's key feature is the automatic partitioning of messages maybe we can update the test to verify messages are produced to different partitions. I was thinking we can  specify a `KAFKA_CREATE_TOPICS` in `docker-compose.yml` to include a topic with three partitions `KAFKA_CREATE_TOPICS: "MyTopic:3:1"`. Then the test can publish messages to those partitions and verify they are consumed. What do you think?

lightswitch05:
Thanks for merging! I can definitely look into adding that extra test for `HighLevelProducer`. Unfortunately I have not had much free time to code this week. Any interest in adding coveralls.io integration?

austinkelleher:
I agree that adding coveralls would be a good idea.

hyperlink:
Yeah, lets do it! Just need to get @haio to activate it (or make me admin?)

lightswitch05:
Added in #432

PR #429, opened 2016-08-05
0.5.3 release

hyperlink:


pcothenet:
@hyperlink I think I'm still getting the `FailedToRebalanceConsumerError: Exception: NODE_EXISTS[-110]` upon restart.

Will confirm.

hyperlink:
I will take a stab at fixing it if I can reproduce it. :)

pcothenet:
Seems to happen for me when I kill the consumers in a non-graceful way, then restart a bunch of them. A few seconds after the new ones start, some of them fail to rebalance.

(Note: I'm trying to avoid non-graceful shutdown but Elastic Beanstalk doesn't play very nice)

hyperlink:
That error means there's a zookeeper node already there. But I am not sure which node it's talking about.

So when you instantiate your consumer are you providing an `id` to the consumer or omitting it? Omitting the `id` will allow the module to generate one based on your groupId + uuid. This is what we do with our consumers and we're not seeing this issue.

pcothenet:
If I do this, it will restart from offset 0 when your consumer starts, no?
Or do you fetch the latest offsets first and start from there?

hyperlink:
offsets are saved based on the group topic and partition so it should be safe.

Issue #467, opened 2016-09-14
Implement Kafka's broker based consumer groups


hyperlink:
PR is up!

PR #477, opened 2016-10-04
Implement Group coordinator

hyperlink:
This PR is implementation of broker coordinator based consumer client for Kafka v0.9+ 

Issue: #467 

CC: @haio @estliberitas


lib/consumerGroup.js
```diff
@@ -0,0 +1,455 @@
...
+      checkProtocol(protocol);
+    }
+
+    protocol.subscription = this.topics;
+    return protocol;
+  }, this);
+};
+
+function checkProtocol (protocol) {
+  assert(protocol, 'protocol is null');
```
andrewaustin: protocol could be another false value and not just null.

lib/consumerGroup.js
```diff
@@ -0,0 +1,455 @@
...
+    } else {
+      checkProtocol(protocol);
+    }
+
+    protocol.subscription = this.topics;
+    return protocol;
+  }, this);
+};
+
+function checkProtocol (protocol) {
```
andrewaustin: Typescript does all these checks for you at compile time ;)

lib/consumerGroup.js
```diff
@@ -0,0 +1,455 @@
...
+  assert(protocol, 'protocol is null');
+  assert(protocol.assign, 'assign function is not defined in the protocol');
+  assert(protocol.name, 'name must be given to protocol');
+  assert(protocol.version >= 0, 'version must be >= 0');
+}
+
+ConsumerGroup.prototype.setCoordinatorId = function (coordinatorId) {
+  this.client.coordinatorId = String(coordinatorId);
```
andrewaustin: Are you sure you want a String object instead of ```this.client.coordinatorId = "" + coordinatorId;```?

lib/assignment/roundrobin.js
```diff
@@ -0,0 +1,74 @@
...
+
+  var ret = _.map(assignment, function (value, key) {
+    var ret = {};
+    ret.memberId = key;
+    ret.topicPartitions = groupPartitionsByTopic(value);
+    ret.version = VERSION;
+    return ret;
+  });
+
+  callback(null, ret);
```
andrewaustin: It doesn't look like any of the code here is async, but you are using callbacks, and elsewhere in the code you treat this code as async. In order to be consistent, should this be wrapped in a setImmediate? (same for the other provided assignment schema.)

lib/consumerGroup.js
```diff
@@ -0,0 +1,455 @@
...
+    function (metadataResponse, callback) {
+      var metadata = mapTopicToPartitions(metadataResponse[1].metadata);
+      debug('mapTopicToPartitions', metadata);
+      callback(null, metadata);
+    },
+
+    function (metadata, callback) {
+      protocol.assign(metadata, groupMembers, callback);
+    }
+  ], callback);
```
andrewaustin: Is it possible that the waterfall method returns both and error and some of the data from protocol.assign? Maybe you should explicitly catch the error from the waterfall, check it and either ```return callback(null, data)``` or just ```return callback(error)``` to prevent any data leaks.

lib/consumerGroup.js
```diff
@@ -0,0 +1,455 @@
...
+    // assign partitions
+    return this.assignPartitions(joinGroupResponse.groupProtocol, joinGroupResponse.members, callback);
+  }
+  callback(null, groupAssignment);
+};
+
+ConsumerGroup.prototype.saveLatestOffsets = function (topicPartitionList, callback) {
+  var self = this;
+  const offsetPayload = _(topicPartitionList).cloneDeep().map(function (tp) {
+    tp.time = -1;
```
andrewaustin: Maybe a comment explaining why you are setting this to -1.

lib/consumerGroup.js
```diff
@@ -0,0 +1,455 @@
...
+  if (this.isLeader) {
+    // assign partitions
+    return this.assignPartitions(joinGroupResponse.groupProtocol, joinGroupResponse.members, callback);
+  }
+  callback(null, groupAssignment);
+};
+
+ConsumerGroup.prototype.saveLatestOffsets = function (topicPartitionList, callback) {
+  var self = this;
+  const offsetPayload = _(topicPartitionList).cloneDeep().map(function (tp) {
```
andrewaustin: Why do a cloneDeep if you immediately call .map()?

lib/consumerGroup.js
```diff
@@ -0,0 +1,455 @@
...
+            if (self.migrator) {
+              offset = self.migrator.getOffset(p, offset);
+            }
+          }
+          p.offset = offset;
+          return p;
+        });
+        callback(null, true);
+      }
+    ], callback);
```
andrewaustin: Same comment from above about leaving errors and data from the waterfall.

lib/consumerGroup.js
```diff
@@ -0,0 +1,455 @@
...
+ConsumerGroup.prototype.getOffset = function () {
+  if (this.offset) {
+    return this.offset;
+  }
+  this.offset = new Offset(this.client);
+  return this.offset;
+};
+
+ConsumerGroup.prototype.connect = function () {
+  if (this.connecting) {
```
andrewaustin: Should this be an error instead?

lib/consumerGroup.js
```diff
@@ -0,0 +1,455 @@
...
+
+ConsumerGroup.prototype.stopHeartbeats = function () {
+  this.hearbeatInterval && clearInterval(this.hearbeatInterval);
+};
+
+ConsumerGroup.prototype.leaveGroup = function (callback) {
+  debug('%s leaving group', this.client.clientId);
+  var self = this;
+  this.stopHeartbeats();
+  if (self.generationId != null) {
```
andrewaustin: Can simplify to ```if (this.generationId)```

lib/consumerGroup.js
```diff
@@ -0,0 +1,455 @@
...
+    cb = force;
+    force = false;
+  }
+
+  if (this.committing && !force) return cb(null, 'Offset committing');
+
+  this.committing = true;
+  setTimeout(function () {
+    this.committing = false;
+  }.bind(this), this.options.autoCommitIntervalMs);
```
andrewaustin: Sometimes you use bind and other times you use ```var self = this```. Should probably pick a method to be consistent to use throughout.

lib/consumerGroup.js
```diff
@@ -0,0 +1,455 @@
...
+    }
+    debug('%s 💚 <-', self.client.clientId, error);
+  });
+};
+
+ConsumerGroup.prototype.fetchOffset = function (payloads, cb) {
+  this.client.sendOffsetFetchV1Request(this.options.groupId, payloads, cb);
+};
+
+ConsumerGroup.prototype.autoCommit = function (force, cb) {
```
andrewaustin: Consider making force not an optional argument.

lib/consumerGroup.js
```diff
@@ -0,0 +1,455 @@
...
+    this.committing = false;
+  }.bind(this), this.options.autoCommitIntervalMs);
+
+  var commits = this.topicPayloads.filter(function (p) { return p.offset !== 0; });
+
+  if (commits.length && this.generationId && this.memberId) {
+    this.client.sendOffsetCommitV2Request(this.options.groupId, this.generationId,
+      this.memberId, commits, cb);
+  } else {
+    cb(null, 'Nothing to be committed');
```
andrewaustin: Not 100% familiar with the API here, but why is the data a string?

lib/consumerGroup.js
```diff
@@ -0,0 +1,455 @@
...
+  }
+};
+
+ConsumerGroup.prototype.commit = ConsumerGroup.prototype.autoCommit;
+
+ConsumerGroup.prototype.close = function (force, cb) {
```
andrewaustin: Another function where you can consider making force a required argument.

lib/consumerGroupMigrator.js
```diff
@@ -0,0 +1,126 @@
...
+  debug('%s connecting consumer group', this.client.clientId);
+  const self = this;
+  if (this.client.ready) {
+    this.consumerGroup.connect();
+  } else {
+    this.client.once('ready', function () {
+      self.consumerGroup.connect();
+    });
+  }
+  this.zk && this.zk.close();
```
andrewaustin: It's not clear why you call ```this.zk.close()``` here.

lib/consumerGroupMigrator.js
```diff
@@ -0,0 +1,126 @@
...
+ConsumerGroupMigrator.prototype.filterByExistingZkTopics = function (callback) {
+  const self = this;
+  const path = '/consumers/' + this.consumerGroup.options.groupId + '/owners/';
+
+  async.filter(this.consumerGroup.topics, function (topic, callback) {
+    const topicPath = path + topic;
+    debug('%s checking zk path %s', self.client.clientId, topicPath);
+    self.zk.exists(topicPath, function (error, stat) {
+      if (error) {
+        return callback(false);
```
andrewaustin: Why ignore errors here?

lib/assignment/range.js
```diff
@@ -0,0 +1,81 @@
+'use strict';
+
+const debug = require('debug')('kafka-node:Range');
```
andrewaustin: I think it's confusing to use virtually no es6 features except const just because node implemented const in old versions. Can you be more explicit about which versions you support in the readme?

lib/protocol/protocol.js
```diff
@@ -533,16 +631,407 @@ function decodeOffsetResponse (resp) {
...
+  return {
+    subscription: topics,
+    version: parsed.version,
+    userData: parsed.userData
+  };
+}
+
 exports.encodeFetchRequest = encodeFetchRequest;
 exports.decodeFetchResponse = decodeFetchResponse;
 exports.encodeOffsetCommitRequest = encodeOffsetCommitRequest;
```
andrewaustin: Rename to ```exports.encodeOffsetCommitRequest``` to be consistent in the naming?

test/assignment/test.range.js
```diff
@@ -0,0 +1,136 @@
...
+      consumer1.topicPartitions['RebalanceTest'].should.eql([0, 1]);
+      consumer1.topicPartitions['RebalanceTopic'].should.eql([0, 1]);
+
+      const consumer2 = _.last(result);
+      consumer2.memberId.should.eql('consumer2');
+      Object.keys(consumer2.topicPartitions).should.eql(['RebalanceTopic', 'RebalanceTest']);
+      consumer2.topicPartitions['RebalanceTest'].should.eql([2]);
+      consumer2.topicPartitions['RebalanceTopic'].should.eql([2]);
+
+      done(error);
```
andrewaustin: Should you explicitly check the error earlier in the test to make sure the test failure message you see is the most helpful one? (Same for other tests...)

lib/assignment/range.js
hyperlink: Planning to follow node TLS support. So 0.12 for a bit longer.

lib/consumerGroup.js
hyperlink: it's only an object if you use `new`

lib/consumerGroup.js
hyperlink: I haven't seen this condition yet. Just added this as a guard.

lib/consumerGroup.js
hyperlink: The result is being consumed by another async.waterfall so it should be caught there I would imagine.

lib/consumerGroup.js
hyperlink: `0` is a valid generation id.

lib/consumerGroup.js
hyperlink: making `commit` method same api as `highLevelConsumer`.

lib/consumerGroup.js
andrewaustin: Can it be undefined?

test/assignment/test.range.js
hyperlink: range and roundrobin are both hardcoded to yield null errors.

andrewaustin: This response doesn't make sense to me. Why pass it to done at the end?

lib/consumerGroup.js
hyperlink: b/c I'm mutating the items.

lib/consumerGroup.js
hyperlink: > Specify -1 to receive the latest offset (i.e. the offset of the next coming message)

lib/protocol/protocol.js
hyperlink: name look consistent. huh?

lib/consumerGroupMigrator.js
hyperlink: `async.filter` doesn't accept error argument.

lib/consumerGroupMigrator.js
hyperlink: Once you join the new consumer group the migrator's job is done and there's no need to keep the zookeeper connection open.

lib/assignment/roundrobin.js
hyperlink: I made the API is async but these two do not need to be.

lib/consumerGroup.js
hyperlink: yes it can be undefined when it tries to leave the group but haven't joined.

lib/consumerGroupMigrator.js
andrewaustin: Restated another way, why are you not throwing an exception here?

andrewaustin: To recap what we talked about in person: If there is an error that causes async.filter self.zk.exists to always fail, you may want to abort with the error rather than continuing with an empty list.

PR #485, opened 2016-10-17
for...in loops must check for hasOwnProperty

Alkarex:
`for...in` is dangerous in JavaScript. When used, one must use the hasOwnProperty() test:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Iterating_over_own_properties_only

Example: kafka-node kept crashing when used together with another library that defined a polyfill for `Object.values()` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values:

``` javascript
if (!Object.values) {
    Object.prototype.values = function (obj) {
        return Object.keys(obj).map(function (key) {
            return obj[key];
        });
    };
}
```

All the `for .. in` loops in kafka-node would iterate over an additional property `values`, which will then make kafka-node to crash.

Note: I tried to run the tests as documented (including installing Docker), but the procedure did not seem to immediately work on Ubuntu 16.04; so I count on Travis...

### DIFF
--- a/lib/assignment/range.js
+++ b/lib/assignment/range.js
@@ -14,6 +14,9 @@ function assignRange (topicPartition, groupMembers, callback) {
 
   const topicMemberMap = topicToMemberMap(groupMembers);
   for (var topic in topicMemberMap) {
+    if (!topicMemberMap.hasOwnProperty(topic)) {
+      continue;
+    }
     debug('For topic %s', topic);
     topicMemberMap[topic].sort();
     debug('   members: ', topicMemberMap[topic]);
--- a/lib/client.js
+++ b/lib/client.js
@@ -524,6 +524,9 @@ Client.prototype.sendToBroker = function (payloads, encoder, decoder, cb) {
     cb = wrap(payloads, cb);
   }
   for (var leader in payloads) {
+    if (!payloads.hasOwnProperty(leader)) {
+      continue;
+    }
     var correlationId = this.nextId();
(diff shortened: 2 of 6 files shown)


hyperlink:
Hi, thanks for the PR.

Curiously, what library implements their polyfill like that? The polyfill links seems to have method set on `Object` instead of `Object.prototype` which seems awfully dangerous.

lib/assignment/range.js
```diff
@@ -14,6 +14,9 @@ function assignRange (topicPartition, groupMembers, callback) {
 
   const topicMemberMap = topicToMemberMap(groupMembers);
   for (var topic in topicMemberMap) {
+    if (!topicMemberMap.hasOwnProperty(topic)) {
```
hyperlink: We can avoid this check if we update `topicMemberMap` instead and used `Object.create(null)` as the initiator to the `_.reduce`.

Alkarex: I have applied the same fix everywhere, but it can probably be optimized in some cases

Alkarex:
@hyperlink It was an internal, non-public library, which I have also updated, so it is fixed on both sides :-)

lib/assignment/range.js
hyperlink: for those other cases I think using lodash's `_.forOwn` might be cleaner.

hyperlink:
Thanks for your contribution!

PR #492, opened 2016-10-24
Kafka topics should not contain special characters

codeforgeek:
This is my observation while playing with createTopics function. I later realised that topic names which I am creating contains special characters such as :,// etc which is causing the issue. I tried without them and it works. It does work with other number and string combinations so no test required for that.

Hope this PR is help.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -357,7 +357,7 @@ Client.prototype.createTopics = function (topics, isAsync, cb) {
   if (typeof isAsync === 'function' && typeof cb === 'undefined') {
     cb = isAsync;
     isAsync = true;
-  }  
+  }
   var validTopics = true;
   Object.keys(topics).map(function (elem) {
     if (!isValid(topics[elem])) {
@@ -365,7 +365,7 @@ Client.prototype.createTopics = function (topics, isAsync, cb) {
     }
   });
 
-  if(!validTopics) {
+  if (!validTopics) {
     return cb(null, 'Some topics contains special characters, please remove them');
   }
   var self = this;
--- a/lib/client.js
+++ b/lib/client.js
@@ -356,11 +356,8 @@ Client.prototype.createTopics = function (topics, isAsync, cb) {
     cb = isAsync;
     isAsync = true;
   }
-  
   validateKafkaTopics(topics);
-  
   var self = this;
-
   // first, load metadata to create topics
   this.loadMetadataForTopics(topics, function (err, resp) {
     if (err) return cb(err);
--- a/lib/client.js
+++ b/lib/client.js
@@ -20,7 +20,7 @@ var Zookeeper = zk.Zookeeper;
 var url = require('url');
 var debug = require('debug')('kafka-node:Client');
 var validateConfig = require('./utils').validateConfig;
-var validateKafkaTopics = require('./utils').validateCreateTopics;
+var validateKafkaTopics = require('./utils').validateTopicNames;
 
 /**
  * Communicates with kafka brokers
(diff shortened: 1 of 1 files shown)


lib/client.js
```diff
@@ -349,12 +349,25 @@ Client.prototype.loadMetadataForTopics = function (topics, cb) {
...
   }
+  var validTopics = true;
+  Object.keys(topics).map(function (elem) {
+    if (!isValid(topics[elem])) {
+      validTopics = false;
+    }
+  });
+
+  if (!validTopics) {
+    return cb(null, 'Some topics contains special characters, please remove them');
```
hyperlink: Lets continue with the error first style and call the callback with a new `Error` as the first argument.

lib/client.js
```diff
@@ -349,12 +349,25 @@ Client.prototype.loadMetadataForTopics = function (topics, cb) {
...
+    return !/[~`!#$%\^&*+=\-\[\]\\';,/{}|\\":<>\?]/g.test(str);
+  }
   topics = typeof topics === 'string' ? [topics] : topics;
 
   if (typeof isAsync === 'function' && typeof cb === 'undefined') {
     cb = isAsync;
     isAsync = true;
   }
+  var validTopics = true;
+  Object.keys(topics).map(function (elem) {
```
hyperlink: `Array.some` could be used here and may not need to iterate through the entire list.

lib/client.js
```diff
@@ -349,12 +349,25 @@ Client.prototype.loadMetadataForTopics = function (topics, cb) {
 };
 
 Client.prototype.createTopics = function (topics, isAsync, cb) {
+  function isValid (str) {
```
codeforgeek: Sure. Make sense.

lib/client.js
codeforgeek: Good one (Y)

lib/client.js
codeforgeek: Cool.

codeforgeek:
@hyperlink I did made the change and created a function in utils. To pass this build, you need to merge the pull no 17.  Sorry for confusion.

lib/client.js
```diff
@@ -355,8 +356,8 @@ Client.prototype.createTopics = function (topics, isAsync, cb) {
     cb = isAsync;
     isAsync = true;
   }
+  validateKafkaTopics(topics);
```
hyperlink: Just realized something. if this fails validation and `isAsync` is true then user will never be notified of the error through the callback. I think we should change this to account for that case.

codeforgeek: I used Error() object with throw keyword in the `validateKafkaTopics ` topics function. So even though isAsync is true, error will be thrown.

hyperlink: When users call it as async method they will expect to catch errors in the callback. So this will end up being an uncaught exception which is unexpected. We should wrap it in a try catch and if it's `isAsync` is true call the callback with the error otherwise throw it again.

codeforgeek: Cool. Make sense. Updating it.

lib/client.js
```diff
@@ -356,7 +356,12 @@ Client.prototype.createTopics = function (topics, isAsync, cb) {
     cb = isAsync;
     isAsync = true;
   }
-  validateKafkaTopics(topics);
+  try {
+    validateKafkaTopics(topics);
+  } catch (e) {
+    if (isAsync) return cb(e);
+    throw new e;
```
hyperlink: `validateTopicNames` throws a new instance so there's no need to `new` the error here. Looks good otherwise.

hyperlink:
Thanks.

PR #520, opened 2016-11-18
Use uuid package

SimenB:
To avoid `warning node-uuid@1.4.7: use uuid module instead`


hyperlink:
Thanks for your contribution!

Issue #541, opened 2016-12-23
HighLevelConsumer leaving partitions without owner on multiple consumers startup


cesaraugustogarcia:
Hi,

when I startup several consumers normally my queue ends up with several partitions without owner (checking with `kafka-consumer-offset-checker.sh `).

In my scenario, I have more than double number of consumers than partitions. For example, 38 consumers for 16 partitions on my queue. And when starting (or restarting) all my consumers I end up always with partitions left with no owner.

hyperlink:
Could you go into more detail on your issue was fixed by exposing the retry configs?

cesaraugustogarcia:
Absolutely. Upon my investigation on startup of my multiple servers, I have serious suspects that the problem is because of contention of servers trying to get ownership of partitions at the same time.

I also noticed that with the default configuration, the period used to retry is always the same, so I focused on having each consumer using a different period of time for the retries, set up when it starts (this mean this period will remain the same, but should be different for each consumer). I was able to get this with this config:
```
rebalanceRetry: {
  retries:    20,
  factor:     1,
  minTimeout: Math.random() * 20000,
  maxTimeout: 20000 + 1,
  randomize:  false
}
```
And the results so far are really good.

cesaraugustogarcia:
@hyperlink Should I close the issue? Or you prefer to leave it open for more feedback?

PR #542, opened 2016-12-23
Allow configuration of rebalance retry strategy on HighLevelConsumer

cesaraugustogarcia:
- This proposal is to address https://github.com/SOHU-Co/kafka-node/issues/541 (HighLevelConsumer leaving partitions without owner on multiple consumers startup). After doing several tests and observations, I believe this (bad) resulting scenario may be caused by race conditions of several consumers trying to consume same partitions and library not handling correctly the retries. Therefore the idea is to allow the rebalance retries to be configurable for each scenario. Particularly, I got good results when setting each consumer with a different (randomised) cycle of retry, to avoid collisions on attempts.
  - The change is not breaking at all.
  - Also is included a bump on retry library (used on rebalancing action) to version 0.7.0 which introduces a couple of bugfixes. The most important fix introduced is https://github.com/tim-kos/node-retry/issues/12

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "nested-error-stacks": "^2.0.0",
     "node-zookeeper-client": "~0.2.2",
     "optional": "^0.1.3",
-    "retry": "~0.6.1",
+    "retry": "^0.10.1",
     "uuid": "^3.0.0"
   },
   "engines": {
--- a/lib/highLevelConsumer.js
+++ b/lib/highLevelConsumer.js
@@ -22,7 +22,14 @@ var DEFAULTS = {
   fetchMinBytes: 1,
   fetchMaxBytes: 1024 * 1024,
   maxTickMessages: 1000,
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "nested-error-stacks": "^2.0.0",
     "node-zookeeper-client": "~0.2.2",
     "optional": "^0.1.3",
-    "retry": "~0.6.1",
+    "retry": "~0.7.0",
```
hyperlink: Can we change this to the latest and use a `^` instead?

cesaraugustogarcia: Absolutely. I just bumped to 0.7.0 trying to be conservative. But I have just checked the latest version, `0.10.0` and has no breaking changes.
So updating my fork changing the commit that has that bump on the retry dependency.

cesaraugustogarcia: Alright @hyperlink version changed.

PR #547, opened 2017-01-07
Consumer Group Heartbeat refactor

hyperlink:
At each heartbeat interval check if last heartbeat responded if there was no response yield timeout error and prevent sending another heartbeat request.


lib/consumerGroup.js
```diff
@@ -53,6 +53,15 @@ function ConsumerGroup (memberOptions, topics) {
   const self = this;
   this.options = _.defaults((memberOptions || {}), DEFAULTS);
 
+  if (!this.options.heartbeatInterval) {
+    this.options.heartbeatInterval = Math.floor(this.options.sessionTimeout / 3);
```
andrewaustin: Why do you divide by three?

lib/consumerGroup.js
```diff
@@ -383,11 +385,14 @@ ConsumerGroup.prototype.startHeartbeats = function () {
...
   debug('%s started heartbeats at every %d ms', this.client.clientId, heartbeatIntervalMs);
-  var self = this;
   this.stopHeartbeats();
-  this.sendHeartbeat();
-  this.hearbeatInterval = setInterval(function () {
-    self.sendHeartbeat();
+
+  let heartbeat = this.sendHeartbeat();
+
+  this.hearbeatInterval = setInterval(() => {
```
andrewaustin: Does this interval eventually have to get cleaned up somewhere?

andrewaustin: oh, I found it in stopHeartbeats()

lib/consumerGroup.js
hyperlink: According to offical kafka doc for `heartbeat.interval.ms`

> The expected time between heartbeats to the group coordinator when using Kafka's group management facilities. Heartbeats are used to ensure that the worker's session stays active and to facilitate rebalancing when new members join or leave the group. The value must be set lower than session.timeout.ms, but typically should be set no higher than 1/3 of that value. It can be adjusted even lower to control the expected time for normal rebalances.

PR #568, opened 2017-01-19
reducing autocommit fires when consumer is idle

brunowonka:
in `HighLevelConsumer.updateOffsets` autoCommit was being called
regardless if the connected topics were updated or not, causing
autocommit to be called frequently with the same offsets already in
place


hyperlink:
Thanks for your contribution @brunowonka

PR #572, opened 2017-01-20
Fix bug in Fetch earliest offsets

bkim54:
when offset isnt ready need to call proper method in setTimeout.

### DIFF
--- a/lib/offset.js
+++ b/lib/offset.js
@@ -74,8 +74,7 @@ function fetchOffsets (offset, topics, cb, when) {
   if (!offset.ready) {
     if (when === -1) {
       setTimeout(() => offset.fetchLatestOffsets(topics, cb), 100);
-    }
-    else if (when === -2) {
+    } else if (when === -2) {
       setTimeout(() => offset.fetchEarliestOffsets(topics, cb), 100);
     }
     return;
--- a/lib/offset.js
+++ b/lib/offset.js
@@ -72,8 +72,11 @@ Offset.prototype.fetchEarliestOffsets = function (topics, cb) {
 // private helper
 function fetchOffsets (offset, topics, cb, when) {
   if (!offset.ready) {
-    setTimeout(() => offset.fetchLatestOffsets(topics, cb), 100);
-    return;
+    if (when === -1)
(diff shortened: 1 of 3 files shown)


bkim54:
@hyperlink requested changes made but your tests fail due to the changes
image

hyperlink:
Right, those tests should fail because they rely on the ready var being changed. Can you update the test to emit `ready` from client instead of changing the value? Emitting should also implicitly set the ready to true.

lib/offset.js
```diff
@@ -23,9 +23,7 @@ util.inherits(Offset, events.EventEmitter);
 
 Offset.prototype.fetch = function (payloads, cb) {
   if (!this.ready) {
-    setTimeout(function () {
-      this.fetch(payloads, cb);
-    }.bind(this), 100);
+    this.once('ready', () => this.fetch(payloads, cb).bind(this));
```
hyperlink: We don't need a bind here since we're using fat arrow.

lib/offset.js
```diff
@@ -43,19 +41,15 @@ Offset.prototype.buildPayloads = function (payloads) {
 
 Offset.prototype.commit = function (groupId, payloads, cb) {
   if (!this.ready) {
-    setTimeout(function () {
-      this.commit(groupId, payloads, cb);
-    }.bind(this), 100);
+    this.once('ready', () => this.commit(groupId, payloads, cb).bind(this));
```
hyperlink: ditto here.

lib/offset.js
```diff
@@ -43,19 +41,15 @@ Offset.prototype.buildPayloads = function (payloads) {
...
   }
   this.client.sendOffsetCommitRequest(groupId, this.buildPayloads(payloads), cb);
 };
 
 Offset.prototype.fetchCommits = function (groupId, payloads, cb) {
   if (!this.ready) {
-    setTimeout(function () {
-      this.fetchCommits(groupId, payloads, cb);
-    }.bind(this), 100);
+    this.once('ready', () => this.fetchCommits(groupId, payloads, cb).bind(this));
```
hyperlink: and here.

bkim54:
@hyperlink, good to go?

hyperlink:
Published as 1.3.3.

PR #580, opened 2017-02-01
Producer broker recovery

hyperlink:


BinitaBharati:
I am having the following dependencies in my project:
```
<dependency>
    	<groupId>org.apache.kafka</groupId>
    	<artifactId>kafka-clients</artifactId>
    	<version>0.11.0.0</version>
</dependency>
 <dependency>
    	<groupId>org.apache.kafka</groupId>
    	<artifactId>kafka_2.11</artifactId>
    	<version>0.11.0.0</version>
</dependency>
```
I have a cluster with 3 running kafka brokers, and a topic of 2 partitions and replication factor of 1. I manually brought down the broker wherever one of the 2 partitions was residing. Even after 30 mins, the leader did not get re-assigned to one of the free up and running broker. Also, the leader was set to -1 for all the 30 mins the original broker was down. The client Producer kept trying to produce into the partition onthe shutdown broker, and kept failing for all of 30 mins. Only after , I manually brought up the shut down broker, the leader got re-assigned back to it, and the client Producer started succeeding in sending messages to that particular broker. In which version of kafka should I find this fix please ?

PR #597, opened 2017-02-12
Allow custom logging implementations

### LINKED ISSUE #595: Support custom log implementations
It would be great if this library could be instructed to use user-configurable logging mechanisms. Specifically, this would make integration into larger applications (and their log handling) easier. On top of this, some logging implementations can be turned on/off during the runtime of an application for improved debugging.

Thinking of this goal, the following would be great to have in `kafka-node`:

 1. Set custom logging implementations, e.g.

```javascript
var kafka = require('kafka-node');
kafka.setLoggerProvider(myLoggerProvider);
```

Where `myLoggerProvider` can be called with a name (`string`) to create a new logger. A logger is any object with a `.warn`, `.info` and `.error` function, e.g. console, bunyan, winston or pino loggers.

Example provider for debug logger:

```javascript
var debug = require('debug');

function debugLoggerProvider(name) {
  var logger = debug(name);
  logger = logger.bind(logger);

  return {
    info: logger,
    warn: logger,
    error: logger
  };
}
```

 2. Continue to use `debug` as the default logging implementation for full backward compatibility.
 3. Adapt the `kafka-node` code to use the user configured logger with reasonable log levels.

The change would be backward compatible and I am willing to contribute it. So if you agree, I'll get right to it.

bripkens:
Kafka-node should be easier to integrate into larger applications. A
large part of this is log handling. Debug's log output configuration is
not really usable for application developers and does not allow
filtering of log messages by level.

This PR adds the ability to set logger providers while retaining the
current default logging strategy: debug.

Fixes #595

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # kafka-node CHANGELOG
 
+## Unreleased
+- Logging strategy should be configurable so that `kafka-node` can be integrated into applications more easily.
+
 ## 2017-02-01, Version 1.3.4
 - Producers should better recover from brokers going offline and coming back [#580](https://github.com/SOHU-Co/kafka-node/pull/580)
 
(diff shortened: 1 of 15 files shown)


bripkens:
I kept the majority of the log statements at the `debug` level as I am not sure how important they truly are. Those statements which log an error have been set to `error`.

hyperlink:
Thanks @bripkens !

PR #598, opened 2017-02-13
Published NPM package should not contain testing and local development artifacts

bripkens:
See: https://docs.npmjs.com/files/package.json#files


hyperlink:
Awesome! Thanks 👍

PR #610, opened 2017-02-22
Passing highWaterOffset both to onMessage and onDone

winstonwp:
This information is useful to keep track of the consumer status.


hyperlink:
Thank you for the contribution!

Issue #619, opened 2017-02-27
ability to have custom partitioner?


bkim54:
can we inject our own partitioner into the HLP?

hyperlink:
I don't think this is possible without code change. Want to submit a PR?

bkim54:
i was able to do it by overwriting the KeyedPartitioner prototype

`KeyedPartitioner.prototype["getPartition"] = getPartition;`

@hyperlink do you think you could approve this PR in DT? https://github.com/DefinitelyTyped/DefinitelyTyped/pull/14911 I needed to declare the KeyedPartitioner in TS

hyperlink:
That seems dangerous. That would affect the implementation for all keyedPartitioner wouldn't it?  It should be easy to update the baseProducer's 'defaultPartitionerType' to accept a constructor function or just the `getPartition` implementation as a regular function.

bkim54:
it would affect all keyedPartitioners yes, but in one application that seems okay? 

anyways above PR creates a new CustomPartitioner type

phreed:
@bkim54 
Is there are reason you made the update in `master` rather than `@types`?
Unless things have changed, that is what is needed in order to get `npm install @types/kafka-node` to work correctly.

PR #625, opened 2017-03-01
Custom partitioner

bkim54:
addresses #619

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Closes the connection to Zookeeper and the brokers so that the node process can 
 * `cb`: **Function**, the callback
 
 ## Producer
-### Producer(client, [options])
+### Producer(client, [options], [customPartitioner])
 * `client`: client which keeps a connection with the Kafka server.
 * `options`: options for producer,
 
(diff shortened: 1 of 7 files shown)


bkim54:
@hyperlink I also updated the typings for kafka-node to reflect the changes in this PR, could you approve that as well? https://github.com/DefinitelyTyped/DefinitelyTyped/pull/14911

Thanks!

bkim54:
@hyperlink wouldnt it also be confusing to override the provided partitioners? 

By providing a CustomPartitionerType the code is more self documenting than using  a `partitionerType` called "default" that uses a user provided getPartition function

hyperlink:
@bkim54 I'm looking at it from the perspective of the module user. So far the user of the module is only exposed to the concepts of partitioner through the `partitionerType`.  With the code as is to add their own implementation requires them remember to change two things versus a single thing of adding a new argument. To me the later is simpler.

bkim54:
why then even provide the concept of `partitionerType`?

Wouldnt it then be better to have one "default partitioner" (a cyclic partitioner is probably the best default) and letting the user pass in a getPartition function

hyperlink:
@bkim54 are you suggesting `partitionerType` could not only take 1, 2, 3 but also a function which would override the built in partitioners?

bkim54:
@hyperlink, no im suggesting providing one default partitioner type and giving users the ability to override the partitioner. (Similar to how the java client works).

hyperlink:
Ok @bkim54 then could we address my initial concern regarding potential user error by throwing an exception if `partitionerType` is not 4 and a custom partitioner function was passed in?

Thanks!

bkim54:
@hyperlink thanks for working through this with me! 😃

hyperlink:
heh np.

PR #654, opened 2017-04-13
Improve performance

crzidea:
`Buffer#slice()` in node is extremely slow and should avoid using it! Unfortunately, we didn't realize it when create this project. So, we should have a plan to improve the performance of this module:

- Step 1: Using `BufferList`(bl) instead of raw Buffer to avoid calling `Buffer#slice()`.
  I have a patch for this module, and tested it in production. In my case, I have reduced CPU usage from **70%** to **50%**, and memory usage from **400m** to **200m**, and increased throughput from **200k/min** to **400k/min**. 
- Step 2: Replace `binary` module. The module is 5 years old and have not be maintained for 5 years! `kafka-node` is heavily depended on `binary` but unfortunately `binary` used `Buffer#slice()`, a lot.

This patch complete `Step 1`,  and I hope someone else could help me with the `Step 2`.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -9,7 +9,6 @@ var async = require('async');
 var retry = require('retry');
 var events = require('events');
 var errors = require('./errors');
-var Binary = require('binary');
 var getCodec = require('./codec');
 var protocol = require('./protocol');
 var BrokerWrapper = require('./wrapper/BrokerWrapper');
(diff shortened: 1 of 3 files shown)


subzero10:
My opinion is that for this type of code you should write something yourself, which you know is fully optimized. This is protocol level and will not be changed often. However, BufferList looks quite good.

hyperlink:
Awesome @crzidea! Looking forward to testing this!

crzidea:
I will Merge this PR and start working on `Step 2`.

crzidea:
@hyperlink The difference in performance would be more convincing if you test on large messages.

PR #667, opened 2017-05-05
Wrong message order (#298)

JesperMadsen123:
* using array to maintain order before decompression

### DIFF
--- a/lib/protocol/protocol.js
+++ b/lib/protocol/protocol.js
@@ -212,7 +212,7 @@ function decodeMessageSet (topic, partition, messageSet, cb, maxTickMessages, hi
                       // errors are logged in offset order too..
                       logger.error('Decompression error causing message loss on topic %s partition %s offset %s - error %s', topic, partition, messageOrderArray[0].offset, messageOrderArray[0].error);
                     }
-                    messageOrderArray = messageOrderArray.shift();
+                    messageOrderArray.shift();
                     ++index;
                   }
                 }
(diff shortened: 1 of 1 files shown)


krystianity:
Did you test this in terms of performance? I just took a quick look and it feels like this could result in additional memory consumption.

JesperMadsen123:
Hi Krystianity, 
Thank you for taking the time to look into this. I saw an extra assignment of the messageorderarray that I removed.

I used the existing tests and did not notice anything out of the ordinary with memory performance, but I would like to improve this patch if there is a problem.

The idea behind this patch is, that when receiving messages, the messages are in order. When decompressing messages asynchronously, the different messages are not guranteed to be finished decompressing in the same order as they where started.  

Therefore the message offsets are stored in an array before decompression, but as soon as decompression is done, and the message is next in offset order, the message is delivered and removed from the array.
The only time messages are stored longer, is when decompression is very much out of sync, with the original order, then the reference is kept until order is restored. E.g. Message 1,2,3,4,5 is received, decompression starts, and finishes as 5,4,3,2,1. In that case the 5 will not be delivered until 1 has been decompressed and has been delivered.

If we take a look at how it normally works, then a compressed chunk of messages has been read, and the library will decompress asynchronously, and will event the message. Since decompression is async, there will be more than one message decompressed in memory at a specific time.

This patch adds an array of objects containing offsets, and if the message is done decompressing out of sequence, the array will store a reference to the message until it can be delivered in order.

To me the only thing that is added memory is that array item with offset and possibly a reference to a decoded message. The array size will be limited to the number of messages that can fit in: fetchMaxBytes.  

Is it the array that makes you concerned with memory performance, or is there something else I have not spotted? 

Is there a specific test scenario you would like me to do? Maybe some unit test or integration test?

PR #677, opened 2017-05-18
Node Ecmascript 5 runtime compatibility

ohollmen:
We're locked into using JavaScript/EcmaScript 5 runtime in our project. Small part of the codebase uses Ecmascript 6 constructs (like let,const,class, new anonymous function style and string interpolation) which prevents interoperability with the stlll mainstream es5 runtime  in older node.js interpreters (bundled in most linux distributions).
The porting effort was very minor and quick to do, hope you have time to review and accept change.
Let me know if you have any questions or if you want me to alter changes to be more compliant kafka-node coding guidelines.
Thanks for the great module !

Olli

### DIFF
--- a/lib/assignment/range.js
+++ b/lib/assignment/range.js
@@ -1,9 +1,9 @@
 'use strict';
 
-const logger = require('../logging')('kafka-node:Range');
-const VERSION = 0;
-const _ = require('lodash');
-const groupPartitionsByTopic = require('../utils').groupPartitionsByTopic;
+var logger = require('../logging')('kafka-node:Range');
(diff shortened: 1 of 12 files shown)


hyperlink:
hey! I worked very hard to add those `const`. Seriously though this feels like a step backward. We shouldn't support versions of node that reached it's end-of-life and both `v0.10` and `v0.12` has ended their support (waaay) back in 2016. The module should strive to keep support in line with Node's [LTS releases](https://github.com/nodejs/LTS#lts-schedule1).

`kafka-node` < 1.0 should work fine those older versions of node.

Issue #632, opened 2017-03-16
ConsumerGroup.addTopics TypeError


zealotds:
## Stack

node_modules/kafka-node/lib/protocol/protocol.js:491
    payloads[topic].forEach(function (p) {
                    ^
TypeError: payloads[topic].forEach is not a function

## More info

kafka-node 1.5.0
kafka 0.10.2.0

hyperlink:
Could you provide some sample code to reproduce this issue? Thanks.

zealotds:
```
var Kafka = require('kafka-node')
var Q = require('q')
var uuid = require('uuid')

var topic = 'test-' + uuid()

Q(new Kafka.HighLevelProducer(new Kafka.Client()))
  .then(producer => {
    var defer = Q.defer()
    producer.on('ready', () => defer.resolve(producer))
    return defer.promise
  })
  .then(producer => {
    console.log('created topic: %s', topic)
    return Q()
      .then(() => Q.ninvoke(producer, 'createTopics', topic))
      .catch(err => console.log('error: %j', err))
      .then(() => producer)
  })
  .then(producer => Q.ninvoke(producer, 'send', [{
    topic: topic,
    messages: [ new Date().toISOString() ]
  }]))
  .then(() => new Kafka.ConsumerGroup({host: 'localhost:2181', groupId: 'test'}, ['test']))
  .tap(consumer => {
    consumer.on('message', msg => console.log('message: %j', msg))
    consumer.on('error', err => console.log('error: %j', err))
  })
  .then(consumer => {
    var defer = Q.defer()
    consumer.addTopics([topic], (err, added) => {
      if (err) return defer.reject(err)
      defer.resolve(added)
    })
    return defer.promise
  })
  .done(added => console.log('added topic: %j: ', added))

```

zealotds:
p.s. if I call the addTopics with "topic" string instead of [topic] array like:
```
    consumer.addTopics(topic, (err, added) => {
      if (err) return defer.reject(err)
      defer.resolve(added)
    })
```
The call would success but returns error:
```
TopicsNotExistError: The topic(s) t,e,s,t,-,6,c,3,0,c,c,b,a,-,8,1,5,d,-,4,7,8,8,-,a,9,b,0,-,7,2,f,3,b,2,8,4,1,d,a,0 do not exist
```
So the topic string here seems be treated like an array of single characters.

michallevin:
Experiencing same issue...

michallevin:
[...]

PR #686, opened 2017-06-14
fix: ConsumerGroup.addTopics TypeError

adoyle-h:
I have tried the `ConsumerGroup.addTopics` in my personal project. The fix works.
`npm test` failed with 3 failings in local. But I think it is unrelated.

image

Fix: #632

### DIFF
--- a/lib/highLevelConsumer.js
+++ b/lib/highLevelConsumer.js
@@ -615,21 +615,38 @@ HighLevelConsumer.prototype.addTopics = function (topics, cb) {
     function (err, added) {
       if (err) return cb && cb(err, added);
 
-      var payloads = self.buildPayloads(topics);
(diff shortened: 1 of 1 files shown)


hyperlink:
Thanks for the PR but I am not sure this will work.  The coordinator and leader of the group should be aware of any new topics added so the ownership of the topic partition can be properly distributed.

Here's what I think you will need the ConsumerGroup to do.

1. Leave group
2. update the `topics` of the ConsumerGroup instance
3. Update the subscriptions of the protocol see `.setupProtocols`
4. join back into the group

There shouldn't be updates to the `HighLevelConsumer` if can be helped. This consumer is getting deprecated.

adoyle-h:
Sorry, I have not fixed the problem for a long time. So I close the PR.

PR #691, opened 2017-06-19
Connect directly to kafka brokers using new client closes #666

### LINKED ISSUE #666: Connect Directly to Kafka Brokers
My company has provided a Kafka environment in AWS, but our DevOps team insists that we should be connecting directly to the Kafka Broker (e.g. 9092). I am able to connect with this approach using kafkacat, but cannot with node-kafka.

The following works...

```
kafkacat -b my-amazon-remote-url:9092 -L
```

But, when I run my code using kafka-node, it does not work.

```
const kafkaNode = require('kafka-node');

const client = new kafkaNode.Client(
 'my-amazon-remote-url:9092',
  'test-client'
);

const producer = new kafkaNode.Producer(client);

producer.on('error', (err) => {
  console.log('NEVER GETS HERE', err);
});

producer.on('ready', () => {
  console.log('NEVER GETS HERE');
});

```

1. Does kafka-node support connecting to the 9092 ports?
2. kafkacat CAN connect this way, so is this just something that has not been implemented in kafka-node or is it simply the wrong way to connect to Kafka? Why is my DevOps team insisting that we use the 9092 port type connection? Would there be a reason?
3. Even when I attempt to connect in the above example using the zookeeper port, I still cannot connect to the remote version.

```
const client = new kafkaNode.Client(
 'my-amazon-remote-url:2181',
  'test-client'
);
...
```

Still no luck, although the above does work on my local Kafka instance (installed on Mac OSX following standard instructions on Kafka.org.)

I appreciate any help on this issue? Thank you!

Kevin

hyperlink:
Closes #666

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,16 @@ Kafka-node is a Node.js client with Zookeeper integration for Apache Kafka 0.8.1
 Follow the [instructions](http://kafka.apache.org/documentation.html#quickstart) on the Kafka wiki to build Kafka 0.8 and get a test broker up and running.
 
 # API
+
+## KafkaClient
+
+New KafkaClient connects directly to Kafka brokers instead of connecting to zookeeper for broker discovery.
+
+### Notable differences
+
+* Constructor accepts an options object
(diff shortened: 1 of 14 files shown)


crzidea:
Good job!

PR #700, opened 2017-07-05
Fix missing key in message sends by Producer issue

### LINKED ISSUE #699: Missing 'key' for message with 'key' field sends by Producer
## Bug Report

In `producer.send(payload)`, if the payload is something like:

```js
{
      topic: 'node-test4',
      messages: 'my test message',
      key: 'lowkey',
}
```

the `key` should be the message key, ref: https://github.com/SOHU-Co/kafka-node#sendpayloads-cb-1

However, the after checking the Kafka side, the key is missing.

### Environment

* Node version: `v8.1.3`
* Kafka-node version: `1.6.2`
* Kafka version: `v0.10.2.1`

* Number of Brokers: 1
* Number partitions for topic: 1

```js
const kafka = require('kafka-node');

const { HighLevelProducer, KeyedMessage, Client } = kafka;

const client = new Client('localhost:2181', 'my-client-id');

client.on('error', error => console.error(error));

const producer = new HighLevelProducer(client, {
  requireAcks: 1,
  partitionerType: 3,
});

let count = 0;
producer.on('ready', () => {
  client.refreshMetadata(['node-test5'], (error) => {
    if (error) {
      console.error(error);
      return;
    }
    setInterval(() => {
      count += 1;
      // Create a new payload
      const payload = [
        {
          topic: 'node-test5',
          messages: `message ${count}`,
          key: 'lowkey',
        },
        {
          topic: 'node-test5',
          messages: new KeyedMessage('highkey', `message ${count}`),
        },
      ];
      producer.send(payload, (err, result) => {
        console.info('Sent payload to Kafka: ', payload);
        if (err) {
          console.error(err);
        } else {
          const formattedResult = result[0];
          console.log(`result: ${formattedResult}`);
        }
      });
    }, 2000);
  });
});

producer.on('error', error => console.error(error));
```

For the comparison, I also send a message which is `KeyedMessage` object.

Here is the output of the `node index.js`

```js
  kafka-node:Client refresh metadata currentAttempt 1 +0ms
Sent payload to Kafka:  [ { topic: 'node-test5',
    messages: 'message 1',
    key: 'lowkey',
    partition: 0,
    attributes: 0 },
  { topic: 'node-test5',
    messages: KeyedMessage { magic: 0, attributes: 0, key: 'highkey', value: 'message 1' },
    partition: 0,
    attributes: 0 } ]
result: undefined
Sent payload to Kafka:  [ { topic: 'node-test5',
    messages: 'message 2',
    key: 'lowkey',
    partition: 0,
    attributes: 0 },
  { topic: 'node-test5',
    messages: KeyedMessage { magic: 0, attributes: 0, key: 'highkey', value: 'message 2' },
    partition: 0,
    attributes: 0 } ]
result: undefined
Sent payload to Kafka:  [ { topic: 'node-test5',
    messages: 'message 3',
    key: 'lowkey',
    partition: 0,
    attributes: 0 },
  { topic: 'node-test5',
    messages: KeyedMessage { magic: 0, attributes: 0, key: 'highkey', value: 'message 3' },
    partition: 0,
    attributes: 0 } ]
result: undefined
```

However, on the Kafka side, the result is different:

```json
[
  {
    "topic": "node-test5",
    "key": "ée",
    "value": "message 1",
    "partition": 0,
    "offset": 0
  },
  {
    "topic": "node-test5",
    "key": "highkey",
    "value": "message 1",
    "partition": 0,
    "offset": 1
  },
  {
    "topic": "node-test5",
    "key": "ée",
    "value": "message 2",
    "partition": 0,
    "offset": 2
  },
  {
    "topic": "node-test5",
    "key": "highkey",
    "value": "message 2",
    "partition": 0,
    "offset": 3
  },
  {
    "topic": "node-test5",
    "key": "ée",
    "value": "message 3",
    "partition": 0,
    "offset": 4
  },
  {
    "topic": "node-test5",
    "key": "highkey",
    "value": "message 3",
    "partition": 0,
    "offset": 5
  },
  {
    "topic": "node-test5",
    "key": "ée",
    "value": "message 1",
    "partition": 0,
    "offset": 6
  },
  {
    "topic": "node-test5",
    "key": "highkey",
    "value": "message 1",
    "partition": 0,
    "offset": 7
  },
  {
    "topic": "node-test5",
    "key": "ée",
    "value": "message 2",
    "partition": 0,
    "offset": 8
  },
  {
    "topic": "node-test5",
    "key": "highkey",
    "value": "message 2",
    "partition": 0,
    "offset": 9
  }
]
```

As you can see, all the `lowkey` key are missing, only key in the `KeyedMessage` are there.
Thanks for your contribution!

twang2218:
fixes #699

Signed-off-by: Tao Wang <twang2218@gmail.com>

### DIFF
--- a/lib/baseProducer.js
+++ b/lib/baseProducer.js
@@ -134,7 +134,7 @@ BaseProducer.prototype.buildPayloads = function (payloads, topicMetadata) {
       if (message instanceof KeyedMessage) {
         return message;
       }
-      return new Message(0, 0, '', message);
+      return new Message(0, 0, _.isEmpty(p.key) ? '' : p.key, message);
     });
 
     let key = p.topic + p.partition;
--- a/test/test.baseProducer.js
+++ b/test/test.baseProducer.js
@@ -48,4 +48,27 @@ describe('BaseProducer', function () {
       done();
     });
   });
+  describe('#buildPayloads', function () {
+    it('should use request "key" as each message key', function () {
+      const fakeClient = new Client();
+
+      fakeClient.topicMetadata = {
+        MyTopic: [0]
+      };
+
+      const producer = new BaseProducer(fakeClient, {}, BaseProducer.PARTITIONER_TYPES.keyed);
+      const payload = [
+        { topic: 'MyTopic', messages: ['hello', 'haloha'], key: 'hikey' },
+        { topic: 'MyTopic', messages: ['message with <undefined> key'] },
+        { topic: 'MyTopic', messages: ['message with <null> key'], key: null },
+        { topic: 'MyTopic', messages: ['message with <true> key'], key: true }
+      ];
+      const messages = producer.buildPayloads(payload, fakeClient.topicMetadata)[0].messages;
+      messages[0].key.should.equal('hikey');
+      messages[1].key.should.equal('hikey');
+      messages[2].key.should.equal('');
+      messages[3].key.should.equal('');
+      messages[4].key.should.equal('');
+    });
+  });
 });


lib/baseProducer.js
```diff
@@ -134,7 +134,7 @@ BaseProducer.prototype.buildPayloads = function (payloads, topicMetadata) {
       if (message instanceof KeyedMessage) {
         return message;
       }
-      return new Message(0, 0, '', message);
+      return new Message(0, 0, p.key || '', message);
```
hyperlink: what happens if the key is number `0`? Would it be better to check if key `!= null` and then use the key otherwise empty string?

hyperlink: hmm looks like that may not make it down the line checkout https://github.com/SOHU-Co/kafka-node/blob/master/lib/protocol/protocol.js#L412

twang2218: Yes, even if I give a pass here, it would still having problem at `encodeMessage()`. So, I guess the `key` must be type of `String`, right? In such case, what if `p.key = true`? I think the boolean value will cause some problem, I think we should use `_.isEmpty(p.key)` to check it.

twang2218:
I updated the PR, now `_.isEmpty()` is used for validating the `p.key`

hyperlink:
Fixed in #704 and published as 2.0.0. Thanks for your help.

PR #708, opened 2017-07-12
"this" should be "self"

sfrooster:
Publishing (KafkaClient) / Consuming (KafkaClient and ConsumerGroup) to a "fresh" broker (no topics), there are two issues:
- I receive "LeaderNotAvailable"
- A null reference exception happens at client.js#541

This fixes the null reference exception.

I still receive "LeaderNotAvailable", followed by "BrokerNotAvailableError: Could not find the leader" and then eventually publishing starting working so there's some timing issue, but I'll look into that separately.

Signed-off-by: Michael Haan <haanmd@gmail.com>

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -279,7 +279,7 @@ Client.prototype.sendOffsetRequest = function (payloads, cb) {
   this.send(payloads, encoder, decoder, cb);
 };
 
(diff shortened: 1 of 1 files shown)


hyperlink:
Thanks for the PR!

hyperlink:
Published as `2.0.1`

PR #722, opened 2017-07-20
use defaultDeep instead of default to merge retryOptions

bkim54:


bkim54:
@hyperlink can you take a look at the test failures? they seem to have nothing to do with the changes i've made.

PR #755, opened 2017-09-04
Flush queued commits of ConsumerGroupStream on tail end of auto-commit time-out

JaapRood:
I ran into a situation where I pulled multiple messages of a `ConsumerGroupStream` in short succession, after which I reached the high water mark. The offset commit was triggered by the first read of the stream, with subsequent commits "throttled" and pushed to a queue. However, because no additional messages were read from the stream, another call to `stream.commit` never happens, with the commit stuck in the queue. This, in turn, lead to messages being consumed more than once.

To solve the issue I added a check to flush anything in the queue at the tail end of the throttling time-out.

I looked for a place to add tests, but couldn't find any for `ConsumerGroupStream` at all. Before I put a lot more effort in I thought I'd check whether this has any chance of getting merged at all! Also happy to take another approach; I'm new to the codebase so not quite sure how this is handled in other areas of it (if at all).

### DIFF
--- a/lib/consumerGroupStream.js
+++ b/lib/consumerGroupStream.js
@@ -81,16 +81,22 @@ class ConsumerGroupStream extends Readable {
       return;
     }
 
+    const commits = convertToCommitPayload(this.commitQueue);
+    this.commitQueued(commits, force, callback);
+  }
+
+  commitQueued (commits, force, callback) {
     this.committing = true;
 
     if (!force) {
       this.autoCommitTimer = setTimeout(() => {
         logger.debug('setting committing to false');
         this.committing = false;
+
+        const queuedCommits = convertToCommitPayload(this.commitQueue);
+        if (!_.isEmpty(queuedCommits)) this.commitQueued(queuedCommits);
       }, this.consumerGroup.options.autoCommitIntervalMs);
     }
-
-    const commits = convertToCommitPayload(this.commitQueue);
 
     if (_.isEmpty(commits)) {
       logger.debug('commit ignored. no commits to make.');


hyperlink:
Sorry. I have failed to add tests in my haste to implement.  @JaapRood that would be great if you could add a test it will have 100% chance of being merged. Thanks for your contribution.

JaapRood:
@hyperlink looks like I won't have the time / attention in the near future to spend on getting the test suite up and running. Feel free to merge without tests (`ConsumerGroupStream` is untested as it is anyway) 😇.

hyperlink:
ok np, I will look into creating a test and merge it soon. :)

fmvilas:
Hey! I need this bug to be fixed as soon as possible so count on me for creating some tests. How do you want to proceed? Shall I create another PR with the content of this one + tests? /cc @hyperlink @JaapRood

hyperlink:
sorry @fmvilas I have dropped the ball on this. I'll merge this now but if you want to throw up a PR for some tests that would be awesome 👍 .

fmvilas:
Thanks! Did you release this change on npm too? I'll try to add tests in the upcoming days.

hyperlink:
No, not yet. There's some kind of travis build error I need to look into.

fmvilas:
Cool. Let me know when it's published, please. Also, if you need some help just raise your hand 😉

hyperlink:
Published as 2.3.0!

fmvilas:
Wow! That was quick! Thank you very much @hyperlink!

Issue #769, opened 2017-09-20
ListGroups and DescribeGroups


wooyeong:
Hi.
 
After upgrading kafka from 0.8 to 0.10, our kafka admin tools didn't show useful numbers, so I am migrating from zookeeper to kafka and trying to implement `fetchCommits`, `listGroups`, and `describeGroups` features to display information on the admin tool properly.
 
I made `ListGroups` and `DescribeGroups`, It seems work well, and I hope kafka-node to have these protocols then our service can utilize the features. I followed instructions on #764, and (manually) tested against 0.9, 0.10, and 0.11.
 
In my opinion, APIs could be located on `kafkaClient` since both protocols are for 0.9+ brokers.
 
For `ListGroups`, I could fetch groupIds from all available brokers by using `client.brokerMetadata` and `client.brokerForLeader`.
 
And for `DescribeGroups`, I think it should accept list of groups as param, but we need to find groups' coordinators to send requests. So I had to use `sendGroupCoordinatorRequest` before asking.
 
What could be done to proceed to pr? Any suggestions?

hyperlink:
Create a PR and I will be glad to review it.

wooyeong:
I'm glad to finish this. Thank you. 😄

PR #770, opened 2017-09-20
ListGroups and DescribeGroups protocol

wooyeong:
#769


wooyeong:
Oh, I think correlationId and request object should go inside mapValuesSeries before this.queueCallback. I'll fix it.

lib/kafkaClient.js
```diff
@@ -413,6 +413,81 @@ KafkaClient.prototype.getApiVersions = function (broker, cb) {
...
+      broker.write(request);
+    }, (err, results) => {
+      if (err) {
+        callback(err);
+        return;
+      }
+      // removing {error: null} before merging
+      results = _.map(results, result => _.omitBy(result, _.isNull));
+
+      callback(null, _.merge({}, ...(_.values(results))));
```
hyperlink: We are still supporting node 4 so you can't use spread operators here.

lib/kafkaClient.js
```diff
@@ -413,6 +413,81 @@ KafkaClient.prototype.getApiVersions = function (broker, cb) {
...
+        return callback(new errors.BrokerNotAvailableError('Broker not available (getListGroups)'));
+      }
+
+      const correlationId = this.nextId();
+      const request = protocol.encodeListGroups(this.clientId, correlationId);
+
+      this.queueCallback(broker.socket, correlationId, [
+        protocol.decodeListGroups,
+        cb,
+        this.options.versions.requestTimeout
```
hyperlink: this can be omitted if using the default `requestTimeout` value.

lib/kafkaClient.js
```diff
@@ -413,6 +413,81 @@ KafkaClient.prototype.getApiVersions = function (broker, cb) {
   broker.write(request);
 };
 
+KafkaClient.prototype.getListGroups = function (callback) {
+  if (!this.ready) {
+    return callback(new Error('Client is not ready (getListGroups)'));
+  }
+  const brokers = this.brokerMetadata;
+  async.mapValuesSeries(brokers,
```
hyperlink: Series is not need here since there's no no ordering requirement on the result correct? Could we change this to `mapValuesLimit` instead with a default (configurable) limit of 10.

lib/kafkaClient.js
```diff
@@ -413,6 +413,81 @@ KafkaClient.prototype.getApiVersions = function (broker, cb) {
...
+      callback(null, _.merge({}, ...(_.values(results))));
+    }
+  );
+};
+
+KafkaClient.prototype.getDescribeGroups = function (groups, callback) {
+  if (!this.ready) {
+    return callback(new Error('Client is not ready (getDescribeGroups)'));
+  }
+  const self = this;
```
hyperlink: Is setting `self` necessary since we're using fat arrows throughout?

lib/kafkaClient.js
```diff
@@ -413,6 +413,81 @@ KafkaClient.prototype.getApiVersions = function (broker, cb) {
...
+        cb(err || null, coordinator ? coordinator.coordinatorId : undefined);
+      });
+    },
+    (err, results) => {
+      if (err) {
+        callback(err);
+        return;
+      }
+
+      async.mapValuesSeries(results,
```
hyperlink: `mapValuesLimit` ?

lib/kafkaClient.js
```diff
@@ -413,6 +413,81 @@ KafkaClient.prototype.getApiVersions = function (broker, cb) {
...
+            return callback(new errors.BrokerNotAvailableError('Broker not available (getDescribeGroups)'));
+          }
+
+          const correlationId = this.nextId();
+          const request = protocol.encodeDescribeGroups(this.clientId, correlationId, groups);
+
+          this.queueCallback(broker.socket, correlationId, [
+            protocol.decodeDescribeGroups,
+            cb,
+            this.options.versions.requestTimeout
```
hyperlink: same comment as above this is optional.

lib/kafkaClient.js
wooyeong: I'll change it to use `.apply` instead

wooyeong:
I've changed all mentioned problems, and introduced a new option `options.maxAsyncRequests` which defaults to 10 to feed `mapValuesLimit`.

For test, I think I can't write tests right now because I have at least minimum working code and have  to make service first. I may try in the 1st week of October if it's not done.

lib/kafkaClient.js
```diff
@@ -413,6 +414,70 @@ KafkaClient.prototype.getApiVersions = function (broker, cb) {
...
+KafkaClient.prototype.getListGroups = function (callback) {
+  if (!this.ready) {
+    return callback(new Error('Client is not ready (getListGroups)'));
+  }
+  const brokers = this.brokerMetadata;
+  async.mapValuesLimit(brokers, this.options.maxAsyncRequests,
+    (brokerMetadata, brokerId, cb) => {
+      const broker = this.brokerForLeader(brokerId);
+      if (!broker || !broker.isConnected()) {
+        return callback(new errors.BrokerNotAvailableError('Broker not available (getListGroups)'));
```
hyperlink: `callback` should be `cb` here.

lib/kafkaClient.js
```diff
@@ -413,6 +414,70 @@ KafkaClient.prototype.getApiVersions = function (broker, cb) {
...
+    }
+  );
+};
+
+KafkaClient.prototype.getDescribeGroups = function (groups, callback) {
+  if (!this.ready) {
+    return callback(new Error('Client is not ready (getDescribeGroups)'));
+  }
+
+  async.groupBy(groups,
```
hyperlink: Could we use `groupByLimit` here instead?

lib/kafkaClient.js
```diff
@@ -413,6 +414,70 @@ KafkaClient.prototype.getApiVersions = function (broker, cb) {
...
+      if (err) {
+        callback(err);
+        return;
+      }
+
+      async.mapValuesLimit(results, this.options.maxAsyncRequests,
+        (groups, coordinator, cb) => {
+          const broker = this.brokerForLeader(coordinator);
+          if (!broker || !broker.isConnected()) {
+            return callback(new errors.BrokerNotAvailableError('Broker not available (getDescribeGroups)'));
```
hyperlink: this should be `cb` here and not that main callback.

lib/kafkaClient.js
```diff
@@ -413,6 +414,70 @@ KafkaClient.prototype.getApiVersions = function (broker, cb) {
...
+          }
+
+          const correlationId = this.nextId();
+          const request = protocol.encodeDescribeGroups(this.clientId, correlationId, groups);
+          this.queueCallback(broker.socket, correlationId, [protocol.decodeDescribeGroups, cb]);
+          broker.write(request);
+        },
+        (err, res) => {
+          callback(err || null, res);
```
hyperlink: could we flatten the result so it's keyed by `groupId`?

wooyeong:
Changed as you requested. I also wrote new admin wrapper class and README.

For `Admin`, I reject if user provides older `Client`, but I don't know it's recommended way to do so.

PR #776, opened 2017-10-07
Added missing callbacks in the commit()

vigneshnrfs:
Added missing callback function in the code. 

This is needed when stream.close() function is called and subsequent clean up happens.

### DIFF
--- a/lib/consumerGroupStream.js
+++ b/lib/consumerGroupStream.js
@@ -78,7 +78,7 @@ class ConsumerGroupStream extends Readable {
 
     if (this.committing && !force) {
       logger.debug('skipping committing');
-      return callback && callback();
+      return callback && callback(null);
     }
 
     this.committing = true;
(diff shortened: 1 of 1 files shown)


lib/consumerGroupStream.js
```diff
@@ -78,7 +78,7 @@ class ConsumerGroupStream extends Readable {
 
     if (this.committing && !force) {
       logger.debug('skipping committing');
-      return;
+      return callback && callback();
```
hyperlink: can we explicitly pass in `null` arg here?

lib/consumerGroupStream.js
```diff
@@ -94,7 +94,7 @@ class ConsumerGroupStream extends Readable {
 
     if (_.isEmpty(commits)) {
       logger.debug('commit ignored. no commits to make.');
-      return;
+      return callback && callback();
```
hyperlink: here as well.

vigneshnrfs:
@hyperlink Made the changes. Thanks.

hyperlink:
Thanks again.

PR #797, opened 2017-11-08
Don't hang process on shutdown

aikar:
The setTimeout will hold a process open during process shutdown, even after the consumer has been closed.

By simply adding unref, node will no longer hold the process open for this task and allow it to shutdown correctly.

### DIFF
--- a/lib/highLevelConsumer.js
+++ b/lib/highLevelConsumer.js
@@ -553,7 +553,7 @@ function autoCommit (force, cb) {
   this.committing = true;
   setTimeout(function () {
     this.committing = false;
-  }.bind(this), this.options.autoCommitIntervalMs);
+  }.bind(this), this.options.autoCommitIntervalMs).unref();
 
   var commits = this.topicPayloads.filter(function (p) { return p.offset !== -1; });
 


aikar:
@hyperlink If you mean closing the consumer, then I'm not seeing any benefit to doing that..

Sure a close operation should clean up the timer, it's not going to really gain anything here.

Yes, it will hold the process open for that window. So not end of the world, but very annoying in development mode and unnecessarily inflates restart times.

I would propose accepting this PR until another reason is found to accomplish this through timer management. Managing the timer will add more complexity for no benefit.

aikar:
@hyperlink any update to this? im currently having to maintain a fork for it and would love if we could just use the mainline package

aikar:
@hyperlink also checking in on this. I still strongly feel unref is the correct approach to dealing with daemon timers such as this. I can't see any cons to it vs the additional maintenance burden managing another timer.

hyperlink:
Hi @aikar I have an idea I wanted to try.  I will work on it after I improve the stability of tests.

aikar:
@hyperlink this change solves the problem cleanly, can we please get this merged?

Issue #820, opened 2017-12-04
highWaterOffset is not included when compression is on


ecksun:
## Bug Report

This seems to have been an issue since highWaterOffset was made part of the API in 938db087ad2edc138e57ec7b0a94a50f0c4ff2dc

Looking at [protocol.js](https://github.com/SOHU-Co/kafka-node/blob/7426cf9c2d6aaeaba9a95f2bcbd75ea7804c3c5b/lib/protocol/protocol.js#L216) we can clearly see `decodeMessageSet` being called without `highWaterOffset` thus leaving it undefined.

### Environment

* Node version: 8.9.1-1nodesource1
* Kafka-node version: 2.2.2
* Kafka version: 0.8.1

### Include Sample Code to reproduce behavior

https://github.com/SOHU-Co/kafka-node/pull/821

PR #821, opened 2017-12-04
Include highWaterOffset with compression

ecksun:
Should solve issue #820 

Its probably possible to write better tests, happy for hints/suggestions on how to most easily do that


hyperlink:
Thank you for your contribution!

Issue #824, opened 2017-12-18
Producer.send() doesn't always invoke callback if broker goes offline (v2.3.0)


thomaslee:
I'm still investigating this one, but wanted to raise it here in case the fix is obvious to you. This bug doesn't _always_ occur & seems a bit timing sensitive, but I can usually reproduce within 1-5 attempts:

1. Start ZK
2. Start Kafka broker
3. Run `DEBUG=kafka-node:* node repro.js` (below)
4. Stop Kafka broker
5. Per DEBUG output, callback queues are cleared without error & `kafka-node:KafkaClient kafka-node-client reconnecting to ...` messages show up on 1s intervals.
6. Start Kafka broker
7. Producer appears to recover, but callback passed to `producer.send(...)` is never invoked & program is "stuck".

Source code for `repro.js`:
```
const kafka = require('kafka-node')
  
function makePayload() {
  return [{
    topic: 'foo',
    messages: ['message']
  }]
}

const client = new kafka.KafkaClient()
const producer = new kafka.Producer(client, {requireAcks: -1})
function producerSendCb(err, data) {
  if (err) {
    console.error(err)
    // work-around for #798
    producer.client.refreshMetadata([], err => {
      if (err) {
        console.error(err)
      }
      send()
    })
    return
  }
  send()
}
function send() {
  producer.send(makePayload(), producerSendCb)
}
producer.on('ready', send)
producer.on('error', err => {
  console.error(err)
})

function cleanup() {
  producer.close(err => {
    if (err) {
      console.error(err)
    }
    process.exit()
  })
}
process.on('SIGINT', cleanup)
process.on('SIGTERM', cleanup)
```

Partial output from `DEBUG=kafka-node:*` (my notes prefixed with `###`):

```
 kafka-node:KafkaClient sending request +0ms
  kafka-node:KafkaClient grouped requests by 1 brokers ["0"] +0ms
  kafka-node:KafkaClient has apiSupport broker is ready +0ms
  kafka-node:KafkaClient Using V2 of produce +0ms
  kafka-node:KafkaClient compressing messages if needed +1ms
  kafka-node:KafkaClient checking payload topic/partitions has leaders +0ms
  kafka-node:KafkaClient found leaders for all +0ms
  kafka-node:KafkaClient sending request +0ms
  kafka-node:KafkaClient grouped requests by 1 brokers ["0"] +0ms
  kafka-node:KafkaClient has apiSupport broker is ready +0ms
[...]

PR #825, opened 2017-12-18
Always clear KafkaClient cb queue with err (#824)

thomaslee:
Failure to do this can cause confusion in user code because callbacks
passed to Producer.send(...) never get invoked.

---

Possible fix for #824, can't reproduce with this change in play. May need some guidance RE: testing.


hyperlink:
Should be fixed in #863 thanks.

PR #836, opened 2018-01-04
Fix the KafkaClient Tests

LeiSun-1101:
Disable some tests from test.kafkaClient.js for Kafka version 0.8.

From the automatic check of this project, it seems the tests are designed for Kafka 0.9+. But from the readme file, the tests should support Kafka 0.8.  This change only fixes for KafkaClient test for Kafka 0.8, this change disables some tests for Kafka 0.8.

### DIFF
--- a/test/test.kafkaClient.js
+++ b/test/test.kafkaClient.js
@@ -109,118 +109,177 @@ describe('Kafka Client', function () {
     });
   });
 
-  describe('Versions', function () {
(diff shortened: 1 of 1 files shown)


LeiSun-1101:
OK. But there are more tests not suitable for Kafka 0.8. I will try to figure all the related tests and disable it if Kafka 0.8 is used.

PR #854, opened 2018-01-17
Update metadatas should extend topicMetadata, not overwrite

rinzeb:


hyperlink:
Hi @rinzeb thanks for the PR. Can you help me understand why we would want to keep an outdated topicMetadata?

thomaslee:
@hyperlink hm I vaguely remember running into something similar but looks like I never got around to opening an issue/PR -- this is probably to play nice with things like `Client.refreshMetadata` which will call `updateMetadatas` with a partial set of topics & blows away metadata for other topics.

For example, `KafkaClient.verifyPayloadsHasLeaders` calls `refreshMetadata` with a partial set of topics based on the set of leaderless payloads. So in this scenario when updateMetadatas is called with an incomplete set of topics, you want to merge the results rather than replace 'em -- otherwise you lose metadata for all the other topics.

Hope that makes sense. Maybe @rinzeb has more to say about it.

rinzeb:
@thomaslee Thanks for your comment, that describes quite accurately what I experienced. Unfortunately I didn't have time to investigate it thoroughly, but in some cases I got 'Leader not available' errors (which weren't always reproducable, sometimes it worked, sometimes it didn't). I tracked it down to the `updateMetadatas` being called with only one of the two topics I was listening on. However, the other topic was already listed in `topicMetadatas`, so I'd thought it would be better to extend the metadata rather than overwriting it. In my case it solved the 'Leader not available' error.

thomaslee:
Right -- sounds very familiar. I can't comment on whether this fix is "correct", but I'd probably have tried something similar. There's certainly _something_ funky going on with `updateMetadatas`. 🙂

hyperlink:
Merged in #857.

PR #857, opened 2018-01-30
updateMetadata should only replace topicMetadata in specific cases

hyperlink:
wrt #854 

cc: @thomaslee @rinzeb


thomaslee:
great 😃

PR #858, opened 2018-01-31
Change metadata require to version 1. This allow client to identify broker rack

spasam:


spasam:
Sorry wrong base branch!

Issue #859, opened 2018-02-02
Support separation of Internal and External tarffic by Broker (KIP-103)


PHameete:
Currently, kafka-node does not have support for brokers that are configured to have separate listeners for internal and external traffic. This was introduced in KIP-103.

There are 2 aspects to this KIP that are important:

1. The protocol of the url's in the endpoints listed in the broker metadata is not necessarily the used protocol. It can instead be the name of the listener which must then be mapped to the protocol via the 'listener_security_protocol_map' metadata that was newly introduced in KIP-103.
2. The kafka-node adapter currently selects the first endpoint that supports the protocol that is selected for usage by the kafka-node adapter. However, there may be multiple endpoints that use this protocol: for example both an INTERNAL as well as an EXTERNAL endpoint. In this case the EXTERNAL endpoint should be chosen as the endpoint that is used because the INTERNAL endpoint can not be reached by the client.

### Reproduction

To reproduce this, set up a Kafka broker that separates INTERNAL and EXTERNAL traffic. You can do this by setting up a Kafka Broker using Docker and using the following configuration:

```
---
version: '3'
services:
  zookeeper:
    image: confluentinc/cp-zookeeper
    hostname: zookeeper
    ports:
      - "3500:3500"
    environment:
      ZOOKEEPER_CLIENT_PORT: 3500
      ZOOKEEPER_TICK_TIME: 2000

  broker:
    image: confluentinc/cp-kafka
    hostname: broker
    depends_on:
      - zookeeper
    ports:
      - "3501:3501"
    environment:
      KAFKA_BROKER_ID: 1
      KAFKA_ZOOKEEPER_CONNECT: 'zookeeper:3500'
      KAFKA_ADVERTISED_LISTENERS: 'INTERNAL://broker:9092,EXTERNAL://localhost:3501'
      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: 'INTERNAL:PLAINTEXT,EXTERNAL:PLAINTEXT'
      KAFKA_LISTENERS: 'PLAINTEXT://0.0.0.0:9092,EXTERNAL://0.0.0.0:3501'
      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
      KAFKA_DEFAULT_REPLICATION_FACTOR: 1
      KAFKA_MESSAGE_MAX_BYTES: 100000000
      KAFKA_REPLICA_FETCH_MAX_BYTES: 100000000
```

[...]

PR #860, opened 2018-02-02
Add support for brokers separating external and internal traffic (introduced in KIP-103)

PHameete:
Fixes point 1 of Issue #859 

Checks if the `listener.security.protocol.map` property is present in the broker metadata. If it is: it will apply the mapping from listener name to security protocol used before checking whether a listener uses the protocol desired by the kafka-node client.

For point 2 it currently still requires the endpoint to be used by kafka-node to be listed first.


hyperlink:
Thanks for your contribution!

PR #889, opened 2018-03-07
auto commit and add event for rebalance

### LINKED ISSUE #787: ConsumerGroup: whenever new consumer is added to group, first few messages consumed twice.
## Questions?
testConsumer.js using  ConsumerGroup to create consumer (auto commit = true, round robin protocol).

Test: - 
 one Topic(testtopic) having 2 partiions. Let prodcuer produce messages over this topic.

Issue:- 
Step 1:- start one testConsumer.js process (lets say Consumer1). Consumer1 will start consuming from both partitions(0 and 1) of the topic (testtopic).
Step 2:- Now start another testConsumer.js process(lets say Consumer2). One of the partitions will be assigned to this topic (lets say 0).
Now observer, consumer1 is still consuming from partitions (0 and 1) for certain time. (heartbeatInterval). on sendHeartbeat() consumer1 will know rebalancing is happening. Post that all rebalancing will happen and it will get one partition (1) and things will run smoothly now.**  

Because of this certain messages (starting messages fetched by  consumer2) are consumed twice. Once by consumer1 and other by consumer2.
Whenever new consumer joins the group the first few messages consumed by new consumer are duplicated(as these messages are already consumed by some other consumer). Is there any way to solve this issue.. ?

My observation:- 
Consumer2 should wait for certain time before fetching messages till consumer1 should register that new consumer has joined and rebalance.
Consumer1 should commit offset forcefully as soon as it knows new consumer has joined.

### Environment

* Node version: 6.10.3
* Kafka-node version: 2.2.3
* Kafka version: 0.11.0.1

* Number of Brokers: 2
* Number partitions for topic: 2

```js
**Currently how consumer1 knows consumer2 has joined. 
ConsumerGroupRecovery.prototype.tryToRecoverFrom = function (error, source) {
// source = heartbeat
  // error.errorCode == 27 is returned when new consumer join
  //reconnect happens
}
```

Thanks for your contribution!

aikar:
Should resolve #787

If auto commit is true, commit before we rejoin the group on a rebalance
According to the kafka protcol, commits are allowed during this stage, as the generation ID has not been bumped yet.

If auto commit is disabled, implementors need a way to decide what to commit, so an onRebalance option was added.

Implementors can perform commits during this method and call back when done to finish the rebalance.

### DIFF
--- a/lib/consumerGroupStream.js
+++ b/lib/consumerGroupStream.js
@@ -50,11 +50,7 @@ class ConsumerGroupStream extends Readable {
           return;
         }
         autoCommitCalled = true;
-        if (self.autoCommit) {
-          self.commit(null, true, callback);
-        } else {
-          callback();
-        }
+        self.commit(null, true, callback);
       };
       if (typeof originalOnRebalance === 'function') {
         try {
--- a/lib/consumerGroupStream.js
+++ b/lib/consumerGroupStream.js
@@ -33,11 +33,40 @@ class ConsumerGroupStream extends Readable {
     super({ objectMode: true, highWaterMark: options.highWaterMark || DEFAULT_HIGH_WATER_MARK });
 
     _.defaultsDeep(options || {}, DEFAULTS);
+    const self = this;
 
     this.autoCommit = options.autoCommit;
 
     options.connectOnReady = false;
     options.autoCommit = false;
+    const originalOnRebalance = options.onRebalance;
+    options.onRebalance = function (isAlreadyMember, callback) {
+      let autoCommitCalled = false;
+      const autoCommit = function () {
+        // We might end up calling this twice due to the try/catch below.
+        // We auto call this on error expecting that autoCommit hasn't been called
+        // yet, but it is possible that it was called, and then an error got thrown
+        if (autoCommitCalled) {
+          return;
+        }
+        autoCommitCalled = true;
+        if (self.autoCommit) {
+          self.commit(null, true, callback);
+        } else {
+          callback();
+        }
+      };
+      if (typeof originalOnRebalance === 'function') {
+        try {
+          originalOnRebalance(isAlreadyMember, autoCommit);
+        } catch (e) {
+          autoCommit();
+          throw e;
+        }
+      } else {
+        autoCommit();
+      }
+    };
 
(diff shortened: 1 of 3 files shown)


aikar:
Depends on #891 to be pulled first, as this work helped surface the off by 1 error fixed in #891

aikar:
This should be good to go now (Travis has been pretty unreliable lately, test did pass)

An example of using the onRebalance event:

```js
    async _onRebalance(isMember: boolean, cb: Function) {
        try {
            if (isMember) {
                this.logger.info("Rebalancing");
                // Clear and wait for any pending event dispatches, giving them a chance to commit
                await this._clearAndWaitForQueue();
                // Now any pending tasks are in the commit queue, which will be picked up after this method finishes
            }
            this.emit("kafka:rebalance");
        } catch (e) {
            this.logger.error("ErrorKafkaRebalance", e);
        } finally {
            cb();
        }
    }
```

aikar:
@hyperlink is anything holding this up?

lib/consumerGroupStream.js
```diff
@@ -33,11 +33,39 @@ class ConsumerGroupStream extends Readable {
...
+    const self = this;
 
     this.autoCommit = options.autoCommit;
 
     options.connectOnReady = false;
     options.autoCommit = false;
+    const originalOnRebalance = options.onRebalance;
+    options.onRebalance = function (isAlreadyMember, callback) {
+      let autoCommitCalled = false;
+      const autoCommit = function (err) {
```
hyperlink: could warp this in `_.once` to avoid having a flag.

aikar: done

hugebdu:
@hyperlink dear Sir, any plans to release this?
10x

hyperlink:
@hugebdu I will get it prepared.

hyperlink:
@hugebdu holding off for now. I'm tracking down an error when running the tests locally.

hugebdu:
@hyperlink 10x for update. waiting patiently

hyperlink:
@hugebdu I found the issue #918

hugebdu:
@hyperlink thanks a lot!

PR #891, opened 2018-03-07
Fix Stale Commit Queue for ConsumerGroupStream

aikar:
The commit at https://github.com/SOHU-Co/kafka-node/commit/abc160986765ebf94b121d56cadaa4a3d426559a introduced a bug
where the offset was incremented at the wrong location.

commitQueued then expected the offset committed to match what is currently in the queue, and that commit broke this.

This results in commit records never being correctly removed from the commitQueue, leading to
weird behavior when committing and messages being consumed twice when rebalances occur relating
to committing stale offsets.

### DIFF
--- a/lib/consumerGroupStream.js
+++ b/lib/consumerGroupStream.js
@@ -19,7 +19,7 @@ function convertToCommitPayload (messages) {
         ret.push({
           topic: topic,
           partition: partition,
-          offset: offset + 1,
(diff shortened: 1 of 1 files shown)


aikar:
@hyperlink have you had a chance to review this? This is a pretty bad bug.

I also have another PR waiting to submit on this as it changes same line.

PR #896, opened 2018-03-12
Bind commit callbacks to when the offset is actually committed.

aikar:
Previously, when using non force commits, the callback was executed
immediately if commit was called while another commit was pending.

This leads to false assurances about that offset actually being successfully
committed to Kafka.

Now, store the callback with the commit request, and wait to execute it
until it is actually committed.


aikar:
@hyperlink is there any concerns with this PR? IMO this is quite important as implementers are currently receiving a false confirmation.

hyperlink:
This is to address the concerns users have when they want to commit manually and need assurances their commit took place? I think it makes more sense to add a method to explicitly commit rather than shoehorn it into the  autoCommit/throttling mechanism.  What do you think?

aikar:
I think there is a misunderstanding of what this fixes.

Right now, the callback parameter passed to commit is useless. It does not give you reliable information.

I expect that when I pass a callback to a commit instruction, regardless of WHEN the commit occurs (instantly, or delayed), that callback should give me the response of whether or not this commit has successfully been recorded in kafka, and has no chance of being re-read again.

Before this PR, it is impossible to get that assurance.

This change makes it so that the callback is fired with the real response back from kafka. I do not see any other reasonable way to do this than this PR, nor do I see any risks with this PR.

Regarding "add a method to explicitly commit" - that API already exists, by using force parameter it does help create that tile, but when using auto commit to improve efficiency through batching, it does not work and this PR solves that.

aikar:
Though, I do see one current issue that this change does not cover.

Say a partition has 3 messages read, and msg 1 commits with batch enabled, and is waiting for a response, then repeat for 2 and 3.
each will overwrite the previous entry, and when it does finally get committed, it'll commit to 3, however the previous callbacks for 1 and 2 will not fire.

easy to fix, so give me a bit to get that in.

To give a real world scenario here. I'm using kafka for an event broker. when I receive a message from kafka, I write a pending job, but do not execute it.

I then commit to kafka, and for my callback, I execute the job on success, or abort the job on failure.

This is very important that the callback fires reliably for me, and only when kafka gives confirmation it received the offset.

If I get 3 messages, and I commit them all, I need all 3 callbacks to fire once that offset is committed.

aikar:
@hyperlink this PR is now ready, the issue i described in march is now resolved.

LOTS of quality improvements here.

aikar:
pretty major issue fixed added now to this PR...
The client was not even passing errors to the callbacks! Every callback was considered successful even if it had an error...

https://github.com/SOHU-Co/kafka-node/pull/896/commits/75ad8d1003649eeac2cedad608d6a93cefa1defe

thynson:
> Say a partition has 3 messages read, and msg 1 commits with batch enabled, and is waiting for a response, then repeat for 2 and 3.
each will overwrite the previous entry, and when it does finally get committed, it'll commit to 3, however the previous callbacks for 1 and 2 will not fire.

It costs me a couple of days to dig this nasty bug out. It prevent me from implementing a reliable consume-commit pipeline. I think it's worth to open a seperated issue to let other people knows its presense.

PR #897, opened 2018-03-12
Option to auto commit offsets on first join - Resolves #886

### LINKED ISSUE #886: Consumer Should auto commit current offset when no saved offset is found
## Bug Report
Given this log: https://gist.github.com/aikar/045590287e581a5528fc27e22ad7b6f3 

I have an example app that immediately on start, connects a consumer and a producer. It publishes a message immediately on start.

Due to timing of this case, the publish is generally happening before the consumer finishes connecting to the broker, causing the offset to be bumped before the group join, meaning that 'offset latest' would exclude that published message.

I can live with this, as its unrealistic for this scenario to happen in real world scenarios, and i can add a delay to the publish.

However, one thing that super confused me, is that why didn't it work when I restarted the process.

The consumer is seeing no saved offset, so it uses latest, but it is not then committing that offset. This treats every restart as if it was a new client if no commit is triggered post join.
If my client joins with a new group ID at offset 4000, I can live with missing those 4000, but I expect to be able to read from 4001+ onward.

So Broken workflow:
```
START
PRODUCER 1
CONSUMER CONNECT [ treated as new, base offset 1]
// missed 1 message - ok with this
STOP

START
PUBLISH 2
CONSUMER CONNECT [ treated as new, base offset 2, expected offset 1]
// missed 2 messages so far
STOP

START
PUBLISH 3
CONSUMER CONNECT [ treated as new, base offset 3, expected offset 1]
// missed 3 messages currently
PUBLISH 4
CONSUME [ consumed offset 4 ]
COMMIT [ committed to offset 4]
STOP

START
PUBLISH 5
CONSUMER CONNECT [ offset 4 ]
CONSUME [ consumed offset 5 ]
COMMIT [ committed to offset 5 ]
STOP
```

I expected after the first `CONSUMER CONNECT [ treated as new, base offset 1]` to be a COMMIT 1

### Environment

* Node version: 8.9.4
* Kafka-node version: 2.4.1
* Kafka version: 0.11 in this case

aikar:
Record offsets at their current position, so that consumers are properly
registered to the topic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -713,7 +713,7 @@ var options = {
   // Offsets to use for new groups other options could be 'earliest' or 'none' (none will emit an error if no offsets were saved)
   // equivalent to Java client's auto.offset.reset
   fromOffset: 'latest', // default
-
+  commitOffsetsOnFirstJoin: true, // on the very first time this consumer group subscribes to a topic, record the offset returned in fromOffset (latest/earliest)
   // how to recover from OutOfRangeOffset error (where save offset is past server retention) accepts same value as fromOffset
   outOfRangeOffset: 'earliest', // default
   migrateHLC: false,    // for details please see Migration section below
--- a/lib/consumerGroup.js
+++ b/lib/consumerGroup.js
@@ -44,6 +44,7 @@ const DEFAULTS = {
   retries: 10,
   retryFactor: 1.8,
   retryMinTimeout: 1000,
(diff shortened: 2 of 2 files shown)


aikar:
@hyperlink is there any concerns with this PR?

aikar:
@hyperlink any update on this?

PR #907, opened 2018-03-23
Clean up the close process of client and consumerGroup

aikar:
If .close is called, we need to ensure that we close cleanly.
Previously, the kafka/ZK client was closed even when it still had
work pending.

This could lead to undesirable outcomes.

Additionally, the callback to .close() was called before everything
completely clean closed too, resulting in a false assurance that
everything was closed cleanly.

Finally, close() may of been called while we were in the middle of
rebalancing. This could cause us to attempt to join/finish joining
a group even though we are closing down (ie: quick process shutdown).

If we detect that a join is pending, we will wait for it to finish
before then sending the leave request.

If for anything during the  connection phase determines we are closing,
skip over every task and rush to the end and leave the group.

This should give us more reliable group membership and quicker,
more reliable, and cleaner shutdown processes.


aikar:
consider this WIP for now until I resolve the tests.

Patafix:
I need this merge request any plan to finish ?

aikar:
@Patafix, well it works for me, but I don't know how to fix the unit tests....

I won't have any time to get to them anytime soon either.

I have a branch https://github.com/aikar/kafka-node/ you can try with all my improvements to this library (note my commits though, snappy was removed!)

I would suggest forking my fork, tweak to as you want, then use your fork as the dep.

PR #918, opened 2018-04-09
Fix issue where a commit could yield a result breaking connect async.waterfall

hyperlink:


aikar:
@hyperlink how is this an issue? the next function doesn't even consume the input.

It would just be discarded. The only thing I see this change doing is dropping whatever the commit responded with and sending null instead.

But since the next function doesn't consume the previous response, how does that change anything?

aikar:
Ohh... I think I see what you are trying to do, but you put this change at the wrong place.

I think you were meaning if the self.options.onRebalance yields a response ,the next auto commit signature would need to be function (foo, callback) { }

Which would make the current callback parameter be in the wrong location.

So yeah I see what the 'issue' is, but this PR put the change in the wrong place.

aikar:
actually hmm i guess that does apply here too. I don't know why I didn't run into this issue myself. 

So this change is needed as you have it, but you also need to apply it to the onRebalance call, as that is 'userland' and undefined on how the implementor will return.

hyperlink:
@aikar yes we should wrap the `onRebalance` callback as well.

aikar:
and conveniently, this issue hit me in production today...!

PR #929, opened 2018-04-12
Validate joinGroupResponse has some required elements

aikar:
If the response is missing key elements, or sends an invalid protocol,
we need to error out. Invalid protocols before was throwing JS errors.

I'm not sure WHY I had this occur, but I had processes crashing in production due to this.


hyperlink:
Curious to know how this occurred.  Maybe something outside modifying `consumerGroup.protocols` or kafka is sent a unexpected JoinGroup response?

aikar:
@hyperlink it happened during some unstable times with our kafka server, so kafka did send some invalid data.

But doesn't hurt to validate and throw error instead of hard crash so figured should add this.

PR #958, opened 2018-04-28
Add create topics support

jlandersen:
This is a first attempt at adding support for the create topics request. I have been wanting this for a while since I work on some "management" tools for Kafka. Most other Kafka clients in other languages also seem to rely on the "metadata" solution.

This supports specifying number of partitions and replication factor. Explicit replica assignments and config entries are left out for now.

A create topic request must be sent to the controller. To find the current controller, this also adds a version 1 metadata request, which includes the controller Id in the response.

There are a couple of things that are less-than-optimal and could use some input for improvement:

- Added a "V2" suffix to resolve the conflict with current versions of `createTopics(..)` and `loadMetadataForTopics(..)`.
- KafkaClient includes these new functions, but the Admin constructor may be more appropriate for create topics, if this is intended to host more admin functionality in the future (would also resolve the `createTopics(..)` conflict on the client)?
- The metadata structure returned from `loadMetadataForTopics(..)` could use a different structure. The first item in the array is an object with all the broker nodes. Second item is an object with a metadata property containing the topics. I've added the `clusterMetadata` property here, to avoid breaking the existing structure. 
- As a follow up to this, I've left the general handling of meta data load/refresh, but the new one could replace the existing one as I see it, so controller id is always available.

Let me know of anything we need to change before this can hopefully get merged in.

Once this is wrapped up, I have delete topics request ready next, and will then add support for config entries on creation of topics.

### DIFF
--- a/lib/kafkaClient.js
+++ b/lib/kafkaClient.js
@@ -241,6 +241,31 @@ KafkaClient.prototype.connectToBroker = function (broker, callback) {
   }, timeout);
 };
 
+KafkaClient.prototype.getController = function (callback) {
+  this.loadMetadataForTopicsV2([], (error, result) => {
+    if (error) {
+      return callback(error);
+    }
+
+    var controllerId = result[1].clusterMetadata.controllerId;
+    var controllerMetadata = result[0][controllerId];
+    var broker = this.getBroker(controllerMetadata.host, controllerMetadata.port);
+
+    if (!broker || !broker.isConnected()) {
+      return callback(new errors.BrokerNotAvailableError('Controller broker not available'));
+    }
+
+    return callback(null, broker);
+  });
+};
+
+KafkaClient.prototype.getBroker = function (host, port, longpolling) {
+  const brokers = this.getBrokers();
+
+  var addr = host + ':' + port;
+  return brokers[addr] || this.setupBroker(host, port, longpolling, brokers);
+};
+
 KafkaClient.prototype.setupBroker = function (host, port, longpolling, brokers) {
   var brokerKey = host + ':' + port;
   brokers[brokerKey] = this.createBroker(host, port, longpolling);
@@ -691,6 +716,29 @@ KafkaClient.prototype.clearCallbackQueue = function (socket, error) {
     });
   }
   delete this.cbqueue[socketId];
+};
+
+// Sends a version 1 metadata request which includes controller node in response
+KafkaClient.prototype.loadMetadataForTopicsV2 = function (topics, cb) {
+  var correlationId = this.nextId();
+  var request = protocol.encodeMetadataV1Request(this.clientId, correlationId, topics);
+  var broker = this.brokerForLeader();
+
+  this.queueCallback(broker.socket, correlationId, [protocol.decodeMetadataV1Response, cb]);
+  broker.write(request);
+};
+
+KafkaClient.prototype.createTopicsV2 = function (topics, cb) {
+  var correlationId = this.nextId();
+  var request = protocol.encodeCreateTopicRequest(this.clientId, correlationId, topics, this.options.requestTimeout);
+  this.getController((error, broker) => {
+    if (error) {
+      return cb(error);
+    }
+
+    this.queueCallback(broker.socket, correlationId, [protocol.decodeCreateTopicResponse, cb]);
+    broker.write(request);
+  });
 };
 
 KafkaClient.prototype.topicExists = function (topics, callback) {
--- a/lib/kafkaClient.js
+++ b/lib/kafkaClient.js
@@ -790,21 +790,21 @@ KafkaClient.prototype.loadMetadataForTopics = function (topics, callback) {
       broker.write(request);
     }
   ], (err, result) => {
-    if (!err) {
-      this.updateMetadatas(result[1]);
-    }
-
     callback(err, result[1]);
   });
 };
 
 /**
  * Creates one or more topics.
- * Requires kafka 0.10 or greater.
  * @param {Array} topics Array of topics with partition and replication factor to create.
  * @param {createTopicsCallback} callback Function to call once operation is completed.
  */
 KafkaClient.prototype.createTopics = function (topics, callback) {
+  // Calls with [string, string, ...] are forwarded to support previous versions
+  if (topics.every(t => typeof t === 'string')) {
+    return Client.prototype.createTopics.apply(this, arguments);
+  }
+
   this.getController((error, broker) => {
     if (error) {
       return callback(error);
--- a/lib/kafkaClient.js
+++ b/lib/kafkaClient.js
@@ -262,6 +262,8 @@ KafkaClient.prototype.getController = function (callback) {
       return callback(new errors.BrokerNotAvailableError('Controller broker not available'));
     }
 
+    this.updateMetadatas(result);
+
     var controllerId = result[1].clusterMetadata.controllerId;
     var controllerMetadata = result[0][controllerId];
 
--- a/lib/kafkaClient.js
+++ b/lib/kafkaClient.js
@@ -15,6 +15,7 @@ const BrokerWrapper = require('./wrapper/BrokerWrapper');
 const errors = require('./errors');
 const validateConfig = require('./utils').validateConfig;
 const TimeoutError = require('./errors/TimeoutError');
+const NotControllerError = require('./errors/NotControllerError');
 const protocol = require('./protocol');
 const protocolVersions = require('./protocol/protocolVersions');
 const baseProtocolVersions = protocolVersions.baseSupport;
@@ -392,6 +393,17 @@ KafkaClient.prototype.setBrokerMetadata = function (brokerMetadata) {
 KafkaClient.prototype.setClusterMetadata = function (clusterMetadata) {
   assert(clusterMetadata, 'clusterMetadata is empty');
   this.clusterMetadata = clusterMetadata;
+};
+
+KafkaClient.prototype.setControllerId = function (controllerId) {
+  if (!this.clusterMetadata) {
+    this.clusterMetadata = {
+      controllerId
+    };
+
+    return;
+  }
+  this.clusterMetadata.controllerId = controllerId;
 };
 
 KafkaClient.prototype.updateMetadatas = function (metadatas, replaceTopicMetadata) {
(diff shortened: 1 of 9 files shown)


jlandersen:
I am looking into the build issue with Kafka 0.9 (metadata version 1 was added in 0.10)

jlandersen:
Updated PR with a few changes to handle clusters of version < 0.10.
Will now return an error if a controller is not found + tests are skipped for version 0.9.

Let me know what you think @hyperlink

lib/protocol/protocol.js
```diff
@@ -332,9 +377,103 @@ function decodeMetadataResponse (resp) {
...
+    .word32bs('correlationId')
+    .word32bs('topicNum')
+    .word16bs('topic')
+    .tap(function (vars) {
+      this.buffer('topic', vars.topic);
+      vars.topic = vars.topic.toString();
+    })
+    .word16bs('errorCode')
+    .tap((vars) => {
+      console.log(vars);
```
hyperlink: This WIP? Could we get rid of this `console.log`?

lib/kafkaClient.js
```diff
@@ -695,6 +726,29 @@ KafkaClient.prototype.clearCallbackQueue = function (socket, error) {
   delete this.cbqueue[socketId];
 };
 
+// Sends a version 1 metadata request which includes controller node in response
+KafkaClient.prototype.loadMetadataForTopicsV2 = function (topics, cb) {
```
hyperlink: We shouldn't continue to create these V{x} version of api calls. It creates unnecessary burden of managing kafka version protocol on the user:

- Override Client `loadMetadataForTopics`
- ensure the broker is ready  and use `getSupportedForRequestType` to get the appropriate version of encoder/decoder  (see example in `sendRequest`)

lib/kafkaClient.js
```diff
@@ -241,6 +241,37 @@ KafkaClient.prototype.connectToBroker = function (broker, callback) {
   }, timeout);
 };
 
+KafkaClient.prototype.getController = function (callback) {
+  this.loadMetadataForTopicsV2([], (error, result) => {
+    if (error) {
+      return callback(error);
+    }
+
+    var controllerId = result[1].clusterMetadata.controllerId;
```
hyperlink: how often does the `controllerId` change?  Would it be safe to cache this on the client so we can avoid doing a metadata request?

lib/kafkaClient.js
```diff
@@ -695,6 +726,29 @@ KafkaClient.prototype.clearCallbackQueue = function (socket, error) {
...
+KafkaClient.prototype.loadMetadataForTopicsV2 = function (topics, cb) {
+  var correlationId = this.nextId();
+  var request = protocol.encodeMetadataV1Request(this.clientId, correlationId, topics);
+  var broker = this.brokerForLeader();
+
+  this.queueCallback(broker.socket, correlationId, [protocol.decodeMetadataV1Response, cb]);
+  broker.write(request);
+};
+
+KafkaClient.prototype.createTopicsV2 = function (topics, cb) {
```
hyperlink: I'm inclined to just call this `createTopics`.  Planning to drop support node 4 so we're making major version change and this can be included in the things that "break".

lib/kafkaClient.js
```diff
@@ -695,6 +726,29 @@ KafkaClient.prototype.clearCallbackQueue = function (socket, error) {
...
+  var correlationId = this.nextId();
+  var request = protocol.encodeCreateTopicRequest(this.clientId, correlationId, topics, this.options.requestTimeout);
```
hyperlink: Could we move the request creation and correlationId a line down inside the handler ?

lib/protocol/protocol.js
```diff
@@ -332,9 +377,103 @@ function decodeMetadataResponse (resp) {
...
+      })
+      .word16bs('errorCode')
+      .word16bs('errorMessage')
+      .tap(function (vars) {
+        if (vars.errorCode === 0) {
+          return;
+        }
+
+        const TIMEOUT_ERROR_CODE = 7;
+        if (vars.errorCode === TIMEOUT_ERROR_CODE) {
```
hyperlink: We should use the error map in the `ERROR_CODE` or just use the `createGroupError` function.

lib/protocol/protocol.js
```diff
@@ -332,9 +377,103 @@ function decodeMetadataResponse (resp) {
...
+  Binary.parse(resp)
+    .word32bs('size')
+    .word32bs('correlationId')
+    .word32bs('topicNum')
+    .word16bs('topic')
+    .tap(function (vars) {
+      this.buffer('topic', vars.topic);
+      vars.topic = vars.topic.toString();
+    })
+    .word16bs('errorCode')
```
hyperlink: WIP? We should handle this error code.

lib/kafkaClient.js
jlandersen: Makes sense - my one concern with this is is that the between api version 0 and >=1 the semantics have changed with regards to the topics parameter.

In version 0 an empty array returns all topics. In version >= 1, a null value (-1) is used to fetch all topics. An empty array returns no topic metadata [1]. This is potentially trouble for clients that use this. I guess this could be put under the next major version umbrella, thus allowing this breaking change. But there are also plenty of calls from inside the lib relying on this (and I think doing something like converting [] -> null would be a bit of a lie for clients).

A suggestion could be:
 - Override it, and make it work with getSupportedForRequestType, which includes the different behaviour depending on version, thus a breaking change similar to createTopics.
 - It is up to clients to be aware of this - basically if they use a version >= 0.10 "null" is now used for fetching all topics.
 - Allow internal use to force version 0 for now to avoid manipulating the input parameter in order to keep all existing functionality that works with < 0.10, running. 

[1] (https://kafka.apache.org/protocol#The_Messages_Metadata).

lib/kafkaClient.js
jlandersen: I have not been able to find documentation that verifies this - my impression is that the election process happens when starting the cluster and otherwise if the current elected broker crashes/becomes unavailable. We can cache it, and update metadata if the request ends up being sent to a broker that isn't controller?

lib/kafkaClient.js
hyperlink: Seems like an ugly change from a users perspective since it puts the burden on the user again to update the behavior of their app depending on the version they're connecting to and would make migrations painful. I think this is a case where we should keep kafka node's behavior consistent and abstract away the ugliness of the kafka  protocol. Calling `loadMetadataForTopics` with empty array should always return all topics.

I always thought fetching all topic metadata seemed inefficient when we're just to get the broker metadata. So perhaps we can have new method called `loadMetadata` that will always pass [] for topic?

What do you think?

jlandersen: So the following changes:

- Change existing `loadMetaDataForTopics` so that if the broker supports api version => 1, then convert an empty array to null value to fetch all topics
- Add new `loadMetadata` that doesen't request for topics (by always passing [] for topic, I assume you mean fetch no topics in this case?). This can then be used in other cases where we just need broker metadata.

This sounds good to me - if you agree i'll try and get all the suggested changes in the PR in sometime this week.

hyperlink: sounds great @jlandersen thanks. 👍

jlandersen:
Bunch of changes are in:

- Removed WIP parts (log, delete topics request encoder/decoder - will come in later PR with the full feature)
- For version 1+ calls of metadata, [] values are sent as null to keep client behaviour of fetching all topics
- Both createTopics and loadMetadataForTopics had their V2 suffix removed, so they replace the original implementations on `Client`.
- I had to add support for existing calls to createTopic - if an array of strings is detected as input, the call is forwarded to original `Client` implementation. Lot's of tests depend on this, so rather than risking blowing it all up with this change, this will keep backwards compatibility for now.
- loadMetadataForTopics on KafkaClient now uses `getSupportedForRequestType` (please check this bit, I added additional logic to make ensure broker is ready and the API version support is resolved).
- Added `loadMetadata` function that refreshes metadata without topics.
- Make sure of `createErrorGroup` when errors are detected in decoder*
- Replaced README part on createTopics with new function documentation (we can also have both version described?)

* note for the createTopics decoder, I distinguish between two types of errors. Overall request errors and individual topic creation errors. Overall request errors are returned as an instance of Error (e.g. timeout or the request somehow got sent to a non-controller). Topic errors are returned as a normal result with the error message provided by Kafka (e.g. setting a replication factor higher than available brokers).

lib/kafkaClient.js
```diff
@@ -695,6 +747,81 @@ KafkaClient.prototype.clearCallbackQueue = function (socket, error) {
...
+
+  this.getController((error, broker) => {
+    if (error) {
+      return callback(error);
+    }
+
+    const correlationId = this.nextId();
+    const request = protocol.encodeCreateTopicRequest(this.clientId, correlationId, topics, this.options.requestTimeout);
+
+    this.queueCallback(broker.socket, correlationId, [protocol.decodeCreateTopicResponse, callback]);
```
hyperlink: Do we want to clear the cached controller createTopics again when we get a `NOT_CONTROLLER` error?

test/test.kafkaClient.js
```diff
@@ -806,4 +806,72 @@ describe('Kafka Client', function () {
...
+          replicationFactor: topic1ReplicationFactor
+        },
+        {
+          topic: topic2,
+          partitions: topic2Partitions,
+          replicationFactor: topic2ReplicationFactor
+        }
+      ], (error, result) => {
+        should.not.exist(error);
+        result.should.be.empty;
```
hyperlink: can we also fetch the topic metadata and verify the partitions/replication factor for newly created topics?

jlandersen: Good point - added.

lib/kafkaClient.js
hyperlink: Good idea. In `Client` we have `sendGroupRequest` so possibly add a `sendControllerRequest`?

jlandersen:
Updated with a `sendControllerRequest` method on `KafkaClient`. It wraps the provided callback with logic that handles a `NotControllerError` error. In this case the cached controller id is removed and the request is redone (based on the `getController` method that will then be forced to fetch latest metadata).

lib/kafkaClient.js
```diff
@@ -999,6 +1004,51 @@ KafkaClient.prototype.verifyPayloadsHasLeaders = function (payloads, callback) {
...
+
+    const originalArgs = _.clone(encoderArgs);
+    const originalCallback = callback;
+    const correlationId = this.nextId();
+    encoderArgs.unshift(this.clientId, correlationId);
+    const request = encoder.apply(null, encoderArgs);
+
+    callback = this.wrapControllerCheckIfNeeded(encoder, decoder, originalArgs, originalCallback);
+
+    this.queueCallback(controller.socket, correlationId, [decoder, callback]);
```
hyperlink: hi @jlandersen thanks for the update.  Could you help me understand why `wrapControllerCheckIfNeeded` is needed? Would it not be simpler to just  inline a function in place of the `callback` that checks if the error is an instance of `NotControllerError` and then set the controller Id to null and call `sendControllerRequest` again if it is?

jlandersen: The reason was to limit the retry to a single time (or potentially x times). If we add a function in place of the callback that always calls the `sendControllerRequest` on a `NotControllerError`, it could keep making requests if the error is returned repeatedly. I am not saying can or will happen, but was rather to be safe. Let me know if you prefer the otherwise :-)

hyperlink:
I'll merge once conflicts are resolved. thanks @jlandersen

jlandersen:
Rebased on master and all checks out! Thanks! Looking forward to adding more things with the controller request in place!

hyperlink:
Published as 3.0.0

PR #971, opened 2018-05-10
Fix error event in recoverFromBrokerChange

thspinto:
# Issue
In ConsumerGroup refresh metadata errors weren't being correctly emitted as 'error'.

# Changes
- Creates an error when metadata refresh request returns an error.
- Fixes emit to send error event to 'error'.


hyperlink:
Published as 3.0.0

PR #977, opened 2018-05-18
Schedules reconnect if fetch isn't started on connection.

thspinto:
I found an edge case that if the consumer connects and the leader for the partition isn't available, the consumer never starts fetching.

You can replicate this problem with a 2 node cluster, and a topic with replication 1. Kill the broker with the partition for the topic, then start the consumer. This will cause the consumer to never start fetching messages.

### DIFF
--- a/lib/consumerGroup.js
+++ b/lib/consumerGroup.js
@@ -532,11 +532,11 @@ ConsumerGroup.prototype.connect = function () {
 
       logger.debug('generationId', self.generationId);
 
-      logger.debug('startFetch is', startFetch);
       if (startFetch) {
         self.clearPendingFetches();
         self.fetch();
       }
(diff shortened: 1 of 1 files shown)


hyperlink:
The consumer can also not fetch if it's not assigned any partitions. Wouldn't this change force these idle consumers to reconnect and disrupt group?

hyperlink:
My concern is if the generation id changes every consumer instance will need to go through the join and sync group flow which is expensive.

When leader isn't available is there an `brokersChanged` event from the client ? There is code to handle this case. Should verify this is working.

thspinto:
@hyperlink Your concern is correct. It does cause the group to resync. I'm going to close this PR and investigate the `brokersChanged`code.

PR #985, opened 2018-05-23
fix listGroups error handling resolves #984

### LINKED ISSUE #984: ListGroups returns errors as successful call
## Questions?

* Check the [Troubleshooting / FAQ](https://github.com/SOHU-Co/kafka-node#troubleshooting--faq) ?
* Search the [issues](https://github.com/SOHU-Co/kafka-node/issues) ?

## Bug Report
When using the `listGroups` method of `KafkaAdmin`, errors are not getting returned properly to the `error` param within the callback. According to this line, any errors that are created will just get set to the `errors` property of `groups`.
https://github.com/SOHU-Co/kafka-node/blob/1c07d7d160b0d407296d7d7bac777f2f7ba6c69f/lib/protocol/protocol.js#L1211-L1230

In the KafkaClient, it doesn't appear that there are any checks for a populated error property. As a result, I get this error on certain clusters within the `res` callback param
https://github.com/SOHU-Co/kafka-node/blob/1c07d7d160b0d407296d7d7bac777f2f7ba6c69f/lib/kafkaClient.js#L463-L466
```
Object {error: Error: ClusterAuthorizationFailed
    at createGroupError
```
Am I incorrect in assuming that this error should be passed down through the `error` param, or is this expected behavior?

### Environment

* Node version: 8.9.3
* Kafka-node version: 2.6.1
* Kafka version: 0.10.2

* Number of Brokers: 3
* Number partitions for topic: 3

```js
const client = new kafka.KafkaClient();
const admin = new kafka.Admin(client); // client must be KafkaClient
admin.listGroups((err, res) => {
  console.log('consumerGroups', res);
});
```

I cannot at this time, but if necessary I can work to include some.

I'm also glad to contribute a PR.

ahmedlafta:
Tested locally via symlinking into another project. I don't see any unit tests for `listGroups` but I can add some if you'd like.

Also, `describeGroups` will likely need the same fix which I'd be glad to submit in a separate PR.

### DIFF
--- a/lib/kafkaClient.js
+++ b/lib/kafkaClient.js
@@ -460,9 +460,7 @@ KafkaClient.prototype.getListGroups = function (callback) {
         callback(err);
         return;
       }
-      // removing {error: null} before merging
-      results = _.values(_.map(results, result => _.omitBy(result, _.isNull)));
-      results.unshift({});
+      results = _.values(results);
       callback(null, _.merge.apply({}, results));
     }
   );
--- a/lib/protocol/protocol.js
+++ b/lib/protocol/protocol.js
@@ -1204,12 +1204,14 @@ function encodeListGroups (clientId, correlationId) {
 
 function decodeListGroups (resp) {
   let groups = {};
+  let error;
+
   Binary.parse(resp)
     .word32bs('size')
     .word32bs('correlationId')
     .word16bs('errorCode')
     .tap(vars => {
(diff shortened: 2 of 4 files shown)


ahmedlafta:
@hyperlink I've added a couple basic tests. I tried avoiding writing too much logic for tests within `kafkaClient`.

PR #1011, opened 2018-06-23
Update client.js

AndyleeSharp:
when "disconnected" event emited from "node-zookeeper-client/lib/ConnectionManager.js", it already begin reconect in its internal [https://github.com/alexguan/node-zookeeper-client/blob/290e4684a1e0b6d824b9e78df00781e131db5861/lib/ConnectionManager.js#L278](url)

if reconnet here(client.js), the zookeeper  TCP Connections will increase one when zookeeper server reboot. for example, first reboot zookeeper server,the num of connection from client to zookeeper is 2; second reboot zookeeper,the num of connection is 3. the kafka connections also increase one per reboot

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -113,13 +113,13 @@ Client.prototype.connect = function () {
       self.emit('error', error);
     }
   });
-  zk.once('disconnected', function () {
+/*  zk.once('disconnected', function () {
     if (!zk.closed) {
       zk.close();
       self.connect();
       self.emit('zkReconnect');
     }
-  });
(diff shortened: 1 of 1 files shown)


hyperlink:
We still need to emit a `zkReconnect` on zookeeper reconnect so the `HighLevelConsumer` can do a rebalance.

AndyleeSharp:
ok, only remove the code of "reconnet" ,remain emit zkReconnect event

hyperlink:
Published as 3.0.0

Issue #999, opened 2018-06-09
ConsumerGroup bug while removeTopics


zzswang:
## Bug Report

consumerGroup.removeTopics will run into crash.

```
/Users/zzs/Workspace/src/github.com/zzswang/watchdog/node_modules/kafka-node/lib/highLevelConsumer.js:814
  this.payloads = this.payloads.filter(function(p) {
                                ^
TypeError: Cannot read property 'filter' of undefined
    at ConsumerGroup.HighLevelConsumer.removeTopics (/Users/zzs/Workspace/src/github.com/zzswang/watchdog/node_modules/kafka-node/lib/highLevelConsumer.js:814:33)
    at /Users/zzs/Workspace/src/github.com/zzswang/watchdog/build/webpack:/src/index.js:54:25
    at process._tickCallback (internal/process/next_tick.js:150:11)
```

### Environment

* Node version: 10
* Kafka-node version: 2.6.1
* Kafka version: 0.11

### Include Sample Code to reproduce behavior

```js
          consumerGroup.removeTopics([topic], () =>
            console.log(`consumer remove topic ${topic}`)
          );
```

### guess reason

consumerGroup call HighLevelConsumer's removeTopics function, 

and in this function, need this.payloads.

it is build in HighLevelConsumer's constructor, but not in ConsumerGroup.

PR #914, opened 2018-04-05
Implement 'addTopics' Functionality in consumerGroup Module

LeiSun-1101:
This is related to https://github.com/SOHU-Co/kafka-node/issues/632

PR #1012, opened 2018-06-26
fix:ConsumerGroup bug while removeTopics

jnotnull:
fix bug: https://github.com/SOHU-Co/kafka-node/issues/999

### DIFF
--- a/lib/consumerGroup.js
+++ b/lib/consumerGroup.js
@@ -591,6 +591,15 @@ ConsumerGroup.prototype.stopHeartbeats = function () {
   this.heartbeatInterval && clearInterval(this.heartbeatInterval);
 };
 
+ConsumerGroup.prototype.removeTopics = function (topics, cb) {
+    topics = typeof topics === 'string' ? [topics] : topics;
+    this.topics = this.topics.filter(function (topic) {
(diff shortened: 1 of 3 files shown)


hyperlink:
`removeTopics` should work similarly to `addTopics` and let the coordinator know. See how PR #914 was implemented.

jnotnull:
Thanks for your comments.

lib/consumerGroup.js
```diff
@@ -660,6 +660,28 @@ ConsumerGroup.prototype.addTopics = function (topics, cb) {
...
+  topics = typeof topics === 'string' ? [topics] : topics;
+
+  async.series([
+    callback => this.client.topicExists(topics, callback),
+    callback => (this.options.autoCommit && this.generationId != null && this.memberId)
+        ? this.commit(true, callback)
+        : callback(null),
+    callback => this.leaveGroup(callback),
+    callback => {
+      this.topics = this.topics.filter(function (topic) {
```
hyperlink: use `_.difference` instead ?

jnotnull: done

hyperlink:
Published as 3.0.0 thanks again.

PR #1017, opened 2018-07-05
Fixing Consumer.addTopics type

bonzzy:
Method addTopics should be Generic because it could be misleading
if you don't know which type is returning callback value of the parameter
added.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -47,7 +47,7 @@ export class Consumer {
   on (eventName: 'message', cb: (message: Message) => any): void;
   on (eventName: 'error' | 'offsetOutOfRange', cb: (error: any) => any): void;
 
-  addTopics (topics: string[] | Topic[], cb: (error: any, added: string[] | Topic[]) => any, fromOffset?: boolean): void;
+  addTopics<T> (topics: T, cb: (error: any, added: T) => any, fromOffset?: boolean): void;
 
   removeTopics (topics: string | string[], cb: (error: any, removed: number) => any): void;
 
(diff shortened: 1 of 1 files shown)


jakagacic:
Great work!

zoltan-nz:
@bonzzy It looks to me the `addTopics` method params accept only array of strings or payload objects.
https://github.com/SOHU-Co/kafka-node/blob/master/lib/consumer.js#L189-L239

The type definition shouldn't give you a hint about accepted params? With generic type, we loose that restriction. What do you think?

bonzzy:
@zoltan-nz Do you think this new changes could be ok?
Generic type must extend `string[] | Topic[]`.

zoltan-nz:
@bonzzy Looks nice.
What would be a good test example for this logic?
We have two options for the previous implementation, but would be great which would test this new extension.
https://github.com/SOHU-Co/kafka-node/blob/master/types/kafka-node-tests.ts#L104-L105

```
consumer.addTopics(['t1', 't2'], (err: any, added: any) => { });
consumer.addTopics([{ topic: 't1', offset: 10 }], (err: any, added: any) => { }, true);
```

hyperlink:
Published as 3.0.0

makhtev:
Maybe implement this  `addTopics` interface in whole library?
There are already some contradictions between documentation and code assosiated with `addTopics` method.

For example in ConsumerStream:
>Similar API as Consumer with some exceptions. Methods like pause and resume in ConsumerStream...

By the doc I expected to see same `addTopics` method logic in `Consumer` and `ConsumerStream`

PR #1022, opened 2018-07-11
Add ConsumerGroupStream typedef

aikar:
update types/node

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "snappy": "^6.0.1"
   },
   "devDependencies": {
-    "@types/node": "^10.0.0",
+    "@types/node": "^10.5.2",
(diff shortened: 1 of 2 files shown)


aikar:
@hyperlink Can we get this pulled please

zoltan-nz:
@aikar Great improvement. Thank you! 

How about extending our TypeScript test with this new interfaces? You can find examples here: https://github.com/SOHU-Co/kafka-node/blob/master/types/kafka-node-tests.ts#L161-L178

What do you think?

mhverbakel:
Shouldn't the `ConsumerGroupStreamOptions` inherit from `ConsumerGroupOptions` instead of `ConsumerOptions`?

PR #1023, opened 2018-07-11
Give more information on why the broker was not found 

aikar:
for BrokerNotAvailable

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -304,9 +304,9 @@ Client.prototype.sendGroupRequest = function (encode, decode, requestArgs) {
   var request = encode.apply(null, requestArgs);
   var broker = this.brokerForLeader(this.coordinatorId);
   var brokerError = null;
-  if (!broker || !broker.socket) {
+  if (!broker) {
     brokerError = 'Could not find broker';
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -303,10 +303,15 @@ Client.prototype.sendGroupRequest = function (encode, decode, requestArgs) {
 
   var request = encode.apply(null, requestArgs);
   var broker = this.brokerForLeader(this.coordinatorId);
-
-  if (!broker || !broker.socket || broker.socket.error || broker.socket.destroyed) {
+  var brokerError = null;
+  if (!broker || !broker.socket) {
```
hyperlink: Do we need this check since the socket should always exist on a broker if it's available?

lib/client.js
```diff
@@ -303,10 +303,15 @@ Client.prototype.sendGroupRequest = function (encode, decode, requestArgs) {
...
+    brokerError = 'Could not find broker';
+  } else if (broker.socket.error || broker.socket.closing || broker.socket.destroyed) {
```
hyperlink: we can use `broker.isConnected`

aikar:
@hyperlink changes nadd

hyperlink:
Published as 3.0.0

Issue #1029, opened 2018-07-16
Producer using Client or KafkaClient does not recover from partition reassignment


david-benes:
When partitions are reassigned from one Broker to another, all following producer attempts to send a message will fail with UnknownTopicOrPartition. The failure continues until partitions are moved back or until the producer is restarted.

This is how I reproduced the issue:

- prepared a cluster of 2 brokers
- created a test topic with 1 partition and replication-factor 1
- prepared a producer script from https://github.com/SOHU-Co/kafka-node/blob/master/example/producer.js
- started producer
- called kafka-reassign-partitions.sh to move the only partition to the next broker

I would expect the client to handle change in partitions. Instead, UnknownTopicOrPartition error is being thrown for all following produce requests (until the partition is moved back or producer restarted).

I tested this with both Client and KafkaClient. The error appeared in both cases.

The symptoms are identical as with #968, however the reason for the error is different. In #968, the issue was caused by special characters. In this case producer just don't update metadata properly.

hyperlink:
Published as 3.0.0

PR #1030, opened 2018-07-16
Fixes an issue when a Producer does not recover from partition reassi…

david-benes:
…gnment (#1029)


hyperlink:
Published as 3.0.0

Issue #1045, opened 2018-08-09
Race condition when closing client, still attempt to refreshBrokerMetadata when client is closing


HansonHH:
### Environment

* Node version: v8.9.0
* Kafka-node version: 2.6.1
* Kafka version: 0.11.0.2

### For specific cases also provide
* Number of Brokers: 3
* Number partitions for topic: 8

### Description

The scenario is, we have 3 Kafka nodes running as a cluster and we restart Kafka consumer groups whenever the scale of Kafka cluster has been changed (remove or add a Kafka broker) or in the case that we need to update SSL certificates for consumer groups. Programmatically, We call close() function of consumer group to shut it down, but sometimes a race condition occurs and we get an assertion error (shown below). I see there is already a merged commit tried to fix this issue (https://github.com/SOHU-Co/kafka-node/pull/942), but the problem seems like ```this.closing``` is ```undefined``` when calling ```refreshBrokerMetadata()``` of KafkaClient and in this case, the condition check in ```refreshBrokerMetadata()``` does not return and continues to call ```getAvailableBroker()```. The ```getAvailableBroker()``` calls ```connectToBrokers()```. In ```connectToBrokers()```, the ```async.doWhilst``` interrupts whenever ```!this.closing && !broker && index < hosts.length``` equals to ```false```. If ```this.closing``` is ```undefined``` before the loop and becomes ```true``` during the loop due to race condition, then ```getAvailableBroker()``` will return a broker to ```loadMetadataFrom()``` which causes the assertion check to fail.

Maybe we should check if ```this.closing``` is set as ```true``` before return a broker?

### Output:
```json
AssertionError: false == true
    at KafkaClient.loadMetadataFrom (/opt/test/node_modules/kafka-node/lib/kafkaClient.js:330:3)
    at async.waterfall (/opt/test/node_modules/kafka-node/lib/kafkaClient.js:315:80)
    at nextTask (/opt/test/node_modules/async/dist/async.js:5324:14)
    at next (/opt/test/node_modules/async/dist/async.js:5331:9)
    at /opt/test/node_modules/async/dist/async.js:969:16
    at async.doWhilst (/opt/test/node_modules/kafka-node/lib/kafkaClient.js:190:16)
    at /opt/test/node_modules/async/dist/async.js:969:16
    at next (/opt/test/node_modules/async/dist/async.js:3022:18)
    at connectToBroker (/opt/test/node_modules/kafka-node/lib/kafkaClient.js:184:9)
[...]

PR #1046, opened 2018-08-09
Return error when connecting to brokers if client is closing

HansonHH:
I've filed an issue regarding this commit, please refer to https://github.com/SOHU-Co/kafka-node/issues/1045. There is a race condition when broker is disconnecting and Kafka client sometimes still attempts to refreshBrokerMetadata when client is closing.

### DIFF
--- a/lib/kafkaClient.js
+++ b/lib/kafkaClient.js
@@ -92,10 +92,13 @@ util.inherits(KafkaClient, Client);
      */
 
 function parseHost (hostString) {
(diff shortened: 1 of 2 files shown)


HansonHH:
Another issue is in the type definitions, the type of key in ProduceRequest should optionally be string or Buffer, because in the case of using schema registry, a message's key or value could be serialized as buffer, so it could be Buffer type.

PR #1050, opened 2018-08-10
Use Map for Callback Queue

aikar:
This is faster than using Objects as maps and will improve performance.


aikar:
looks like the test failures were due to travis instability and not due to my changes

aikar:
@hyperlink can we get this pulled, it's holding progress on other prs

Issue #998, opened 2018-06-08
ConsumerGroupStream offset alway equal to 0


nferno:
## Questions?

* Check the [Troubleshooting / FAQ](https://github.com/SOHU-Co/kafka-node#troubleshooting--faq) ?
* Search the [issues](https://github.com/SOHU-Co/kafka-node/issues) ?

## Bug Report
The offset of Messages that consume from ConsumerGroupStream alway equel to 0

### Environment

* Node version: 8.9.4
* Kafka-node version: 2.6.1
* Kafka version: 1.1.0

### For specific cases also provide
* Number of Brokers: 3
* Number partitions for topic: 100

### Include Sample Code to reproduce behavior

```js
const ConsumerGroupStream = require('kafka-node').ConsumerGroupStream;
const resultProducer = new ProducerStream();

const consumerOptions = {
  kafkaHost: '127.0.0.1:9092',
  groupId: 'TestGroup',
  sessionTimeout: 15000,
  protocol: ['roundrobin'],
  asyncPush: false,
  id: 'consumer',
  fromOffset: 'earliest'
};

const consumerGroup = new ConsumerGroupStream(consumerOptions, 'TestTopic');

const messageTransform = new Transform({
  objectMode: true,
  decodeStrings: true,
  transform (message, encoding, callback) {
    console.log(message);
    callback(null, {
      topic: 'RebalanceTopic',
      messages: `You have been (${message.value}) made an example of`
    });
  }
});

consumerGroup.pipe(messageTransform).pipe(resultProducer);
```

### Include output

```js
{ topic: 'TestTopic',
  value: 'test1',
  offset: 0,
  partition: 4,
  highWaterOffset: 310666,
  key: null,
  timestamp: '2018-05-31T20:32:46.186Z' }
{ topic: 'TestTopic',
  value: 'test2',
  offset: 0,
  partition: 4,
  highWaterOffset: 310666,
  key: null,
  timestamp: '2018-05-31T20:33:06.533Z' }
```

Thanks for your contribution!

daniloaburto:
Same issue here.
Using `host` with zookeeper url instead `kafkaHost` might help.

aikar:
that is odd, can you try with fromOffset: latest?

I also use kafka 1.1.0 with CGS and do not have this issue. I suspect your out of offset range and something is wrong when using earliest.

I do not use Client either, only KafkaClient.

abhinavfaujdar:
@aikar 
[...]

PR #1051, opened 2018-08-17
Skip processing buffer after partial message detected

verakruhliakova:
Kafka server may add partial message to the end of the message set with, for example, the following data:
```
MessageSet => [Offset MessageSize Message]
  Offset => 0
  MessageSize => 16216
Message => Crc MagicByte Attributes Key Value 
  Crc => 0
  MagicByte => 0
  Attributes => 0
  Key => buffer of 0 bits
  Value => buffer of 0 bits
```

Instead of key/value the message is filled with 0 bits sequence. I couldn't find any special requirements on the format of the partial message, so maybe this is allowed behavior. But with such message from the server consumer continues decoding 0 bits, emits empty messages and the worst - it ends up with array of 0 offsets, e.g. [96510, 96511, 96512, 0, 0, 0, 0, 0], which leads to starting reading the next batch from 0 offset and gets into an infinite cycle.

(Took info about protocol from https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-Messagesets and https://kafka.apache.org/documentation.html#impl_reads )

### DIFF
--- a/lib/protocol/protocol.js
+++ b/lib/protocol/protocol.js
@@ -263,12 +263,13 @@ function decodeMessageSet (topic, partition, messageSet, cb, maxTickMessages, hi
   const messageSetSize = messageSet.length;
   while (messageSet.length > 0) {
     var cur = 8 + 4 + 4 + 1 + 1 + 4 + 4;
+    var partial = false;
     Binary.parse(messageSet)
       .word64bs('offset')
       .word32bs('messageSize')
       .tap(function (vars) {
         if (vars.messageSize > messageSet.length - 12) {
-          vars.partial = true;
+          partial = true;
         }
       })
       .word32bs('crc')
@@ -309,7 +310,7 @@ function decodeMessageSet (topic, partition, messageSet, cb, maxTickMessages, hi
           return;
         }
 
-        if (!vars.partial && vars.offset !== null) {
+        if (!partial && vars.offset !== null) {
           messageCount++;
           set.push(vars.offset);
           if (!cb) return;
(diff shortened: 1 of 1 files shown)


aikar:
Could this be what is happening in #998 ?

verakruhliakova:
@aikar I think #998 is different. Here client emits fake empty messages, like 
```
{
  "topic": "events-streaming",
  "value": "",
  "offset": 0,
  "partition": 0,
  "highWaterOffset": 167780,
  "key": ""
}
```
And after this it starts reading from 0 offset again. But the order of messages and offset sequence are correct.
In #998 client emits real messages with 0 offset but with correct values.

ahmedlafta:
FYI I've tested this code change locally against a 2.0 broker and it resolves some issues reported in against this library (e.g. https://github.com/SOHU-Co/kafka-node/issues/1054)

If we could get a release with this code fix that'd be great.

thomaslee:
> If we could get a release with this code fix that'd be great.

+1 to that -- if this is what I think it is, this is a consumer-breaking bug for folks on 2.x. We were bitten by a similar issue in Sarama: https://github.com/Shopify/sarama/pull/1149

praveenkumaresan:
+1 to that. We have tested this code locally against 2.0 broker and it resolves issue we faced consuming empty messages in a loop!

aikar:
Why is kafka sending these messages now anyways?

tvvignesh:
We are facing this bug too. Kindly release this fix as soon as possible. Thanks.

hyperlink:
Published as 3.0.0

Issue #906, opened 2018-03-22
Is there a way to `pipe` results after a message is produced using ProducerStream?


wmartins:
## Questions?

I'm trying to "know" when a given message was produced when using `ProducerStream`. Is there any way to do that?

I'm trying to `pipe` it using the "stream flow", however, I'm not able to get the desired result (code below).

## Bug Report

I'm not able to know when a result was published on Kafka using `ProducerStream`.

### Environment

* Node version: `v8.5.0`
* Kafka-node version: `2.4.1`
* Kafka version: `0.9.0.1`

### For specific cases also provide
* Number of Brokers: `1`
* Number partitions for topic: `5`

### Include Sample Code to reproduce behavior

```js
const { ProducerStream, Client } = require('kafka-node');
const { Transform } = require('stream');

const client = new Client('localhost:2181');

const producer = new ProducerStream({
    kafkaClient: client,
    producer: {
        partitionerType: 2,
    },
});

const transformStream = new Transform({
    objectMode: true,
    transform (chunk, encoding, callback) {
        String(chunk).split('\n').forEach(msg => {
            if (!msg) {
                return;
            }

            console.log('Pushing', msg);

            this.push({
                messages: msg,
                topic: 'topic1',
            });
        });

        callback();
    },
});

function main() {
    process.stdin.pipe(transformStream).pipe(producer).on('data', (data) => {
        console.log('Produced!', data);
    });
}

main();
```

#### To run:

```bash
echo 'A\nB\nC\nD\nE' | node src/producer.js
```

#### Output:

```
Pushing A
Pushing B
Pushing C
Pushing D
Pushing E
```

#### Expected Output:

**NOTE:** The results might be out of order, but something like the folowing output. Also, the output could include things like `topic`, `offset`, `partition` and other.

```
Pushing A
Pushed A
Pushing B
Pushing C
Pushed B
Pushed C
Pushing D
Pushed D
Pushing E
Pushed E
```

### Include output with [Debug](https://github.com/SOHU-Co/kafka-node#how-do-i-debug-an-issue) turned on

```
  kafka-node:KafkaClient Connect attempt 1 +0ms
  kafka-node:KafkaClient Trying to connect to host: 127.0.0.1 port: 9092 +20ms
  kafka-node:KafkaClient Sending versions request to 127.0.0.1:9092 +49ms
  kafka-node:KafkaClient broker socket connected {"host":"127.0.0.1","port":"9092"} +12ms
Pushing A
[...]

PR #1060, opened 2018-09-03
Implement ProducerStream using Duplex Streams

MUI-Pop:
This is regarding #906 , I have implemented it using Duplex stream,


MUI-Pop:
Just noticed the batch/buffer producer wouldnt send result per message.
Its because the result sent by client.sendProduceRequest isn't per message instead just the topic data. 

Any thoughts ?

PR #1063, opened 2018-09-05
V3.0.0

hyperlink:


aikar:
@hyperlink per our PM on discord, I really recommend keeping this 2.X - there was no backwards incompat changes that I'm aware of in latest changes.

We should reserve 3.0 for major API breakage.

bumping major will scare people into thinking its a riskier update than it really is.

aikar:
Oh blah I forgot about the node 4 drop, and I didn't realize createTopics was breaking...

I guess 3.x is necessary, and we will have to use 4.0 for major API changes.

PR #1064, opened 2018-09-05
Don't transmitMessages if already reading

aikar:
This causes fetch requests to be sent out in the middle
a read operation, causing some weird ordering to occur.


aikar:
I can't recall the exact scenario that I was fixing with this here, but I recall seeing order of events emitted on the consumer being really weird, and traced it down here.

this avoids it calling .resume() when 'done' is also going to call it.

If someone understands streams really well can tell me why this is bad, if you think its bad, i'd love to know it.
but this did not break anything for us and fixed something.

Issue #1065, opened 2018-09-05
ConsumerGroupStream: commit with force=true will break future commits with force=false


navbruce:
## Bug Report

Hopefully a reasonably self-evident logic bug:

In consumerGroupStream.js, if `commit(msg, force=true, cb)` is called (with a valid message & offset) it will in turn call `commitQueued(commits, force=true, cb)`.  `commitQueued()` then sets `this.committing=true`, but skips the `setTimeout()` call that is required to set `this.committing=false`.

Any future calls to `commit(msg, force=false, cb)` will bail out on the `if(this.committing && !force)` test, and therefore `commitQueued()` will not be called.

Basically, `this.committing=true` should only happen if the `setTimeout()` call will happen, or else `this.committing` will never be reset to false.

The end result of this bug is that offsets are never being auto-committed on a topic I'm consuming with ConsumerGroupStream.

PR: https://github.com/SOHU-Co/kafka-node/pull/1066

### Environment

* Node version: 8.11.4
* Kafka-node version: 2.6.1, 3.0.0
* Kafka version: 1.1.0

PR #1066, opened 2018-09-05
ConsumerGroupStream: Only set committing=true if it can be set false later

navbruce:
Ensure that calling `commitQueued()` with `force=true` won't unbalance the `this.committing` true/false assignments.

Issue https://github.com/SOHU-Co/kafka-node/issues/1065

### DIFF
--- a/lib/consumerGroupStream.js
+++ b/lib/consumerGroupStream.js
@@ -107,9 +107,9 @@ class ConsumerGroupStream extends Readable {
   }
 
   commitQueued (commits, force, callback) {
-    this.committing = true;
-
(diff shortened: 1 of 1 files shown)


aikar:
This looks like a good temporary fix @hyperlink 
It roughly matches my logic: https://github.com/aikar/kafka-node/commit/a19b3e34e7502a7c9faba776a2e9c280fbbb5cda#diff-591c50c13e2d409f8187f227a2f6c13bR87
in when the boolean should be set

hyperlink:
Thanks for your contribution!

PR #1067, opened 2018-09-06
Use double ended queue instead of Array for message buffer

hyperlink:


aikar:
One concern I do have, is this will be a breaking change if anyone depended on accessing that field.

The project should really use the _ prefix for private fields that shouldn't be touched.

hyperlink:
> One concern I do have, is this will be a breaking change if anyone depended on accessing that field.

imo that field is not documented so it's fair game to be changed.

thomaslee:
Excellent. 👍

PR #1068, opened 2018-09-06
Fix a few omissions in types/index.d.ts

thomaslee:
These type definitions are super handy, but noticed a few of them may be a bit out of date.


thomaslee:
Closing this -- botched the change in haste. I'll reopen when I've got more time to focus on it.

PR #1071, opened 2018-09-07
Partition-aware async consumer proof of concept

thomaslee:
@hyperlink @aikar thought I'd put together something for discussion RE: async message processing in consumers & addressing some of the limitations of `ConsumerGroup` and `ConsumerGroupStream`. This is just a proof of concept so very rough around the edges, it's probably riddled with bugs and if we throw this away in favor of something more suitable that's totally fine. I was hoping to get this up before 3.x shipped, but it's probably too big to squeeze in at the last minute anyway.

The problems I'm trying to solve here are mostly in relation to `ConsumerGroupStream` and `ConsumerGroup` since that's "the future", but apply just as much to the older ZK stuff I think.

Example output from `example/async.js` in case it helps you follow along in the code: https://gist.githubusercontent.com/thomaslee/3f60249e82322f1d7edba57d061c252f/raw/dc2193f6d57e016db961d4744fa0bf16ad3fa3dc/output.txt

## Implemented

1. **Async message processing "for free".** `ConsumerGroup` obviously doesn't support this at all out of the box without some careful finagling. `ConsumerGroupStream` takes care of _some_ of that finagling, but it's still easy for users to make mistakes (e.g. somebody not paying attention could attempt to commit offset 3 after previously committing offset 5).
2. **Parallelism at the partition level "for free".** Slow processing of a message on topic/partition A/1 (e.g. a large file, a slow database/web service, whatever) will not slow down processing on topic/partitions A/2 or B/3.
3. **In-order processing at the partition level "for free".** This one is a bit more subtle: `ConsumerGroupStream` makes it possible, but it's another one that's easy to get wrong if you're not paying attention. In `AsyncConsumerGroup` you need to explicitly `ack()` a message or resolve a promise/`async` method to indicate that you're ready to process the next message on a partition.
4. **_Acquiesce_ and commit before a rebalance/close.** The `acquiesce` methods you see wait for processing to reach a "quiet" state & then waits for pending commits to complete prior to allowing a rebalance to occur. The existing commit logic in `ConsumerGroup` is a bit more haphazard in that it can't really ensure that all processing is complete before rebalances etc. occur. `ConsumerGroupStream` does better, but falls back to `ConsumerGroup` semantics in certain important cases.
5. **Throttling of fetches with watermarks** `ConsumerGroupStream` throttles by batches, which can mean we're waiting for fetches when we could be processing. The watermarks allow us to fetch more aggressively up to some limit (`prefetchHighWatermark`) and then wait for processing to reduce the backlog below some floor value (`prefetchLowWatermark`) before we start fetching again. The result is theoretically a bit less "thrashy" & should ensure less idle time that could be spent processing messages. Even better than this would be the "Partition-aware fetching" described below.

## Not implemented but fairly trivial

1. **Upper bounds on "processed-but-not-yet-committed" messages.** Network issues or cluster outages can prevent consumers from committing offsets. While that's happening, existing consumer implementations will charge forward and process however many messages have been previously fetched. This can lead to large amounts of "reprocessing" when the consumer finally makes contact with the cluster again & goes through a storm of rebalances. It would be fairly trivial to set an upper bound on this per partition, per topic and/or per consumer to minimize reprocessing. Useful for use cases where reprocessing "hurts" & there's a desire to keep it to a minimum. If you want to make sure you're never 1 message ahead of what's been committed & don't mind the throughput hit, that's fine. **(EDIT: now implemented)**
2. **Commit on an interval (`autoCommit`) instead of on completion of processing.** I personally don't think there's a huge problem with just committing immediately after processing a message & relying on the natural throttling that occurs in high volume topics/partitions, but some folks prefer to commit on an interval. This would be easy to wire up in `CommitManager`. My understanding is this is at best difficult to do correctly with the current `ConsumerGroupStream` API.

## Not implemented & difficult (4.x? :wink:)

These typically involve large or tricky changes to KafkaClient/ConsumerGroup internals.

1. **Unified consumer API.**  I'm sort of loathe to add _yet another_ consumer class & feel like longer term we should consolidate around a unified consumer & producer with pluggable "details" (e.g. ZKCommitManager vs. KafkaCommitManager). I think the `consume(cb)` API would work for both synchronous & asynchronous message processing, though `msg.ack()` is a bit meaningless in synchronous processing. Perhaps async processing would be "good enough" for synchronous use cases.
2. **Partition-aware fetching.** Unless you're very lucky, some topic/partitions will be significantly busier (or take more time to process) than others. The `PartitionProcessor`s know when they're about to run out of work & could inform the code that handles fetching that they're going to need more work soon. This ensures that we can keep processing all partitions even if one partition lags behind.

## Roadmap?

If we're okay with adding the extra consumer class, **I'd love to see something like this land in a hypothetical 3.1.0/3.2.0** -- mostly because I'd sleep a lot easier knowing that our internal users aren't reliant on some internal dirty hacks around kafka-node. 😄 But obviously plenty to discuss here.

## Summary

Hopefully you guys can ignore the warts of the actual implementation & see what I'm shooting for here. Kafka's partitions are the key to safely parallelizing message processing & by treading carefully we can maximize async message processing performance, minimize "reprocessing" that can occur due to rebalances or faults, keep Kafka's delivery order guarantees at the partition level in  a way that's trivial for end users.

And again, probably bugs galore and I've probably got certain things hopelessly wrong -- feel free to comment on the code if something's unclear, but I'm more interested in discussing the *spirit* of the change & the broad strokes of the design/API. I'm not attached to the implementation, but I am very attached to addressing the problems it solves. 😄 If we can do all of this with less code & a drastically different API, that's totally fine by me.

## Example usage

Putting this way down the bottom because the API isn't the interesting part of this IMO, but see `example/async.js` for a full example if you'd like something to play with:

```javascript
const consumer = new kafka.AsyncConsumerGroup(options, 'kafka-node-demo');
consumer.on('error', err => ...);

//
// callback is invoked per message, next message processed after msg.ack()
// return a promise & msg.ack() will be implicitly called when the promise resolves.
//
consumer.consume(msg => {
  console.log(JSON.stringify(msg));
  doSomethingAsync().then(() => msg.ack());
});
// later: consumer.close().then(() => process.exit(0));
```

### DIFF
--- a/example/async.js
+++ b/example/async.js
@@ -4,12 +4,14 @@ const kafka = require('..');
 
 const options = {
   kafkaHost: 'localhost:9092',
-  groupId: 'kafka-node-demo'
+  groupId: 'kafka-node-demo',
+  maxUncommittedMessages: 5,
 };
 
 function doSomethingAsync () {
   return new Promise(resolve => {
-    setTimeout(() => resolve(), 50 + Math.random() * 200);
+    // setTimeout(() => resolve(), 50 + Math.random() * 200);
+    resolve();
   });
 }
 
--- a/example/async.js
+++ b/example/async.js
@@ -1,0 +1,46 @@
+'use strict';
+
+const kafka = require('..');
+
+const options = {
+  kafkaHost: 'localhost:9092',
+  groupId: 'kafka-node-demo'
+};
+
+function doSomethingAsync () {
+  return new Promise(resolve => {
+    setTimeout(() => resolve(), 50 + Math.random() * 200);
+  });
+}
+
+const consumer = new kafka.AsyncConsumerGroup(options, 'kafka-node-demo');
+consumer.on('error', err => {
+  console.error('error: unexpected error from consumer', err);
+});
+
+//
+// `consumer.consume(async msg => ...)` also works, automatically invoking
+// ack() when the promise resolves.
+//
+consumer.consume(msg => {
+  console.log(JSON.stringify(msg));
+
+  //
+  // just demonstrate that we're processing messages asynchronously
+  //
+  doSomethingAsync().then(() => msg.ack());
+});
+
+console.log('waiting for messages to arrive ¯\\_(ツ)_/¯');
+
+function exitGracefully () {
+  console.log('closing consumer');
+  consumer.close()
+    .then(() => {
+      console.log('consumer closed, exiting');
+      process.exit(0);
+    });
+}
+
+process.once('SIGTERM', exitGracefully);
+process.once('SIGINT', exitGracefully);
--- a/kafka.js
+++ b/kafka.js
@@ -1,6 +1,7 @@
 exports.HighLevelConsumer = require('./lib/highLevelConsumer');
 exports.HighLevelProducer = require('./lib/highLevelProducer');
 exports.ProducerStream = require('./lib/producerStream');
+exports.AsyncConsumerGroup = require('./lib/asyncConsumerGroup');
 exports.ConsumerGroup = require('./lib/consumerGroup');
 exports.ConsumerGroupStream = require('./lib/consumerGroupStream');
 exports.Consumer = require('./lib/consumer');
--- a/lib/asyncConsumerGroup/asyncConsumerGroup.js
+++ b/lib/asyncConsumerGroup/asyncConsumerGroup.js
@@ -1,0 +1,269 @@
+'use strict';
+
+const EventEmitter = require('events');
+const ConsumerGroup = require('../consumerGroup');
+const PartitionProcessor = require('./partitionProcessor');
+const CommitManager = require('./commitManager');
+const debug = require('debug')('kafka-node:asyncConsumerGroup:AsyncConsumerGroup');
+
+const DEFAULT_OPTIONS = {
+  maxUncommittedMessages: 1,
+  prefetchHighWatermark: 10000,
+  prefetchLowWatermark: 1000,
+  commitRetryOptions: {
+    minTimeout: 100,
+    maxTimeout: 5000,
+    factor: 2.5,
+    randomize: true
+  },
+  processRetryOptions: {
+    minTimeout: 100,
+    maxTimeout: 5000,
+    factory: 2.5,
+    retries: 5,
+    randomize: true
+  }
+};
+
+//
+// TODO global and per-partition processing stats, commit stats, etc.
+//
+class AsyncConsumerGroup extends EventEmitter {
+  constructor (options, topics) {
+    super();
+
+    if (!options) {
+      options = {};
+    }
+
+    //
+    // TODO if user-provided onRebalance is called, we should invoke it (or perhaps we could offer an alternative API)
+    //
+    const overrides = {
+      paused: true,
+      autoCommit: false,
+      onRebalance: (isMember, cb) => this._onGroupRebalanceBegin(isMember, cb)
+    };
+
+    this.options = Object.assign({}, DEFAULT_OPTIONS, options, overrides);
+    this.closing = false;
+    this.closed = false;
+    this.buffered = 0;
+    this.idle = true;
+    this.consuming = false;
+
+    this.group = new InternalConsumerGroup(this.options, topics);
+    this.group.on('error', err => this._onInternalError(err));
+
+    this.commitManager = new CommitManager(this.group, this.options);
+    this.commitManager.on('error', err => this._onInternalError(err));
+
+    this.partitionProcessors = new Map();
+  }
+
+  consume (cb) {
+    if (this.consuming) {
+      throw new Error('consume() should only be called once');
+    }
+
+    this.consuming = true;
+
+    this.commitManager.on('internal:commit-complete', info => this._onCommitComplete(info));
+    this.commitManager.on('internal:commit-retry', err => this._onCommitRetry(err));
+    this.group.on('rebalanced', () => this._onGroupRebalanceEnd());
+    this.group.on('message', msg => this._onGroupMessage(msg, cb));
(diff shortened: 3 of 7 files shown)


thomaslee:
Oh, and once we can settle on a direction I suppose I'll write some tests. 😅

aikar:
Def a great start, I was looking to do this myself, but totally nice to have someone else take it as I have too much to do anyways..

So far I think design wise you are on the right track.

You may want to check my fork: https://github.com/aikar/kafka-node/ and see the improvements I've made to the commit process for CGS (I use rebase strategy, so can't really link to a specific commit as it keeps changing), which fixes some critical things I'm trying to get merged into master once I figure out some test failures.

Though you really need ensure we factor in support for Transactions. I'm looking to work on transactions before end of the year, and we need to make sure the commit process for the emitted events can support the concept of transactions too.

I'm a bit too busy to review this at this time though, but interested in seeing where you get this.

thomaslee:
I haven't used transactions at all -- but aren't they mostly producer-side? At a glance it seems like the consumer would just need to accommodate the `read_committed` isolation level if it doesn't already. I'm probably missing something here.

aikar:
I haven't reviewed impl yet, but note that ack must return a promise and return result of the commit, as In my process, I operate on a:

1. Receive message
2. Prepare to process message by saving message to disk
3. Commit the message
4. If error, delete the message off disk, if success, execute.

I've also considered a "prepare-ahead" thought process where I could read X messages, prepare them all, commit all X in the same partition (batched with other partitions too), then when result comes back, execute all X or roll all back.

As otherwise, concurrency is limited to partition counts, and prepare-ahead would solve that.

aikar:
I've also mentioned to @hyperlink  that I would really love to see all old kafka classes dropped and only maintain Kafka Broker classes in this client.

This would mean we can name the consumer/producer simply: Consumer, Producer, as I'm not a fan of the Async name used here. this is node, everything's async.

have 1 solid impl for each, with intuitive names that match the java client

Right now i'm constantly struggling on my PR's with test failures on legacy stuff that I can never justify fixing on company time, and I'm too busy with my other massive list of open source projects to do it in my free time.

IMO there is tons of bugs in this projects code, and it's really hard to justify fixing it for deprecated systems, and anyone who insists on using the older versions could and is already likely using the Blizzard client. I would really like to see this client focus on building a solid choice for Kafka based Consumers, and drop ZK support.

Would really love to see this be a target for v4, where we drop a ton of legacy code, come up with a new solid foundation for new consumer/producer code.

I would also like to see the consumer expose a producer that uses the same Client (which should be the KafkaClient renamed)

thomaslee:
> hmm crap, well I thought transactions also had part in the commit process too... if that's the case, then I guess I won't be adding transactions!

😄 do let me know if it turns out I'm wrong about that

> note that ack must return a promise and return result of the commit

It accepts a callback at the moment, but a promise makes more sense. The callback is an implementation detail from our wrapper that snuck over in my "port". But yeah, knowing that a commit has been confirmed is useful info.

> I would really love to see all old ZK based classes dropped and only maintain Kafka Broker classes in this client.
> I'm not a fan of the Async name used here. this is node, everything's async.
> have 1 solid impl for each, with intuitive names that match the java client
> I would really like to see this client focus on building a solid choice for Kafka based Consumers, and drop ZK support.

100% agree on all this. Until then: agree the Async prefix sucks, open to alternatives.

> I would also like to see the consumer expose a producer that uses the same Client

-1 on this idea as stated, primarily because I don't think producers need to be coupled to consumers at all. If you need both a producer & a consumer that share a Client, seems trivial to create 'em that way and just sort of inject them both / pass them in where you need 'em.

thomaslee:
> In my process, I operate on a:
>
> 1. Receive message
> 2. Prepare to process message by saving message to disk
> 3. Commit the message
> 4. If error, delete the message off disk, if success, execute.

Substitute "somewhere safe" with "disk" and this is a very common general pattern, yup.

> I've also considered a "prepare-ahead" thought process where I could read X messages, prepare them all, commit all X in the same partition (batched with other partitions too), then when result comes back, execute all X or roll all back.

In other GH issues I think I talked about "pipelining" which refers to a similar idea. You can get "natural" batching from Kafka by continuing to process messages while you're waiting for earlier offset commits to complete. It looks something like this for any given partition:

1. Process message A, `ack()`, don't wait for commit confirmation
2. Commit of A.offset begins
3. While the commit of A.offset is still in flight, messages B, C, D arrive, get processed and `ack()`ed (again, we don't wait for commit confirmation)
4. Commit of A.offset completes, commit of D.offset begins ...

Then ideally you'd have an upper bound (your "X") on the number of uncommitted offsets such that if something goes wrong with brokers / the consumer and commits start failing or drastically slow down, you guarantee you're never more than X messages ahead of what's committed to Kafka -- either for any given partition or "globally", whatever makes sense for your use case.

The pipelining rubbish fairly well supported by the `AsyncConsumerGroup` and I'd argue the "right" way to do it for anything where throughput is a big deal (the main alternative being `autoCommit`-style interval-based commits). I have no upper bound on X in ACG just yet, though -- this is "Not implemented but fairly trivial" no. 1 😄

aikar:
The reason for producer w/ the consumer is because the consumer is responsible for closing its client.

but now i think about it, could just make producer use consumers client and close consumer to close them both.

thomaslee:
> but now i think about it, could just make producer use consumers client and close consumer to close them both.

Yeah I think if I create a Client and pass it in to a consumer or a producer, responsibility for closing the Client should be mine (i.e. the producer/consumer only implicitly closes the underlying client _if they implicitly created the client_).

FWIW Sarama (Go client) does something similar using a simple boolean flag depending on the constructor you use: https://github.com/Shopify/sarama/blob/master/consumer.go#L98

PR #1072, opened 2018-09-08
Multiple fixes to message ordering and compression, Fixes #298

### LINKED ISSUE #298: Wrong message order

aikar:
There was numerous risks with the fetch process resolved here.

1) Compressed messages, as mentioned in #298, would be guaranteed out of order
We resolve this by making fetch response buffering be handled through a chain
of executions, where one partition is not emitted until the previous one is done,
and able to be processed asynchronously.

This however introduces new risk as the system only dealt with synchronous
processing before, meaning 2 fetch buffers could not be processing at the same
time. But because it's now async, we could end up processing multiple.

We now need to guard starting a new fetch while one is still processing, to ensure
that a new fetch does not begin with stale offset data.

2) Fetch requests may start, and end up being long polling. This fetch result may
end up coming back after a rebalance, or disconnection from the group occurred.

Those fetch requests would then emit events, which may be at different
offsets, topics, or partitions than the new consumer group membership is responsible for...

So if you received an old fetch response for a topic/partition you are no longer
responsible for, you would receive errors. Or even worse, if you are still responsible
for this partition, you may receive a message that about to be received in the new
memberships fetch request.

We will now run a state validator on the message handler to ensure our group
membership has not changed since the fetch request started, and ignore the response
if it has.

3) Kafka, for some unknown reasons, and one known, in the case of compression,
can return an offset older than the offset you requested.

This is a message we know we have already emitted if our topic payloads already
contained this offset.

We need to skip this message, and any message that shouldn't of been in the response
in the first place should we of received an abnormal response, to guarantee data accuracy.

### DIFF
--- a/lib/kafkaClient.js
+++ b/lib/kafkaClient.js
@@ -1128,7 +1128,7 @@ KafkaClient.prototype.sendFetchRequest = function (
   payloads = _.cloneDeep(payloads);
   const memberId = consumer.memberId;
   const generationId = consumer.generationId;
-  function stateValidator (err, type, message) {
+  function stateValidator (unused, type, message) {
     const payloadMap = consumer.payloadMap;
     if (consumer.closing || consumer.connecting || consumer.rebalancing || consumer.memberId !== memberId || consumer.generationId !== generationId) {
       logger.error('ignoring message due to it being from an old group - memberId: ' + memberId, '!=' + consumer.memberId + ' - generationId: ' + generationId + '!=' + consumer.generationId);
--- a/lib/kafkaClient.js
+++ b/lib/kafkaClient.js
@@ -1137,6 +1137,7 @@ KafkaClient.prototype.sendFetchRequest = function (
     if (type === 'message') {
       const {topic, partition, offset} = message;
       if (!payloadMap[topic] || payloadMap[topic][partition] == null) {
+        logger.error('received unexpected message', message, payloadMap);
         // We should have never received this in the first place
         return false;
       }
--- a/lib/kafkaClient.js
+++ b/lib/kafkaClient.js
@@ -1125,6 +1125,30 @@ KafkaClient.prototype.sendFetchRequest = function (
   maxTickMessages,
   callback
 ) {
+  payloads = _.cloneDeep(payloads);
+  const memberId = consumer.memberId;
+  const generationId = consumer.generationId;
+  function stateValidator (err, type, message) {
+    const payloadMap = consumer.payloadMap;
+    if (consumer.closing || consumer.connecting || consumer.rebalancing || consumer.memberId !== memberId || consumer.generationId !== generationId) {
+      logger.error('ignoring message due to it being from an old group - memberId: ' + memberId, '!=' + consumer.memberId + ' - generationId: ' + generationId + '!=' + consumer.generationId);
+      return false;
+    }
+    if (type === 'message') {
+      const {topic, partition, offset} = message;
+      if (!payloadMap[topic] || payloadMap[topic][partition] == null) {
+        // We should have never received this in the first place
+        return false;
+      }
+
+      if (offset == null || offset < payloadMap[topic][partition]) {
+        // Kafka may send an older message than we expect (compressed messages, and other unknown reasons)
+        return false;
+      }
+    }
+
+    return true;
+  }
   if (callback == null) {
     callback = _.noop;
   }
@@ -1141,7 +1165,7 @@ KafkaClient.prototype.sendFetchRequest = function (
           data: {
             payloads: payloads,
             args: [fetchMaxWaitMs, fetchMinBytes],
-            decoderArgs: [this._createMessageHandler(consumer), maxTickMessages]
+            decoderArgs: [this._createMessageHandler(consumer, stateValidator), maxTickMessages]
           }
         };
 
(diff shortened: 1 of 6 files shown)


aikar:
@hyperlink so far tests are passing. 

This is a highly critical fix, that solves numerous issues with this client, that's been the source for so many bug reports.

I commented on quite a few issues linked above that I believe will be resolved with this work.

hyperlink:
Is it worth continuing to support message sets vs implementing the new record batch? I wonder if the ordering issues will go away with record batch.

> But because it's now async, we could end up processing multiple.

We could flip the `waiting` flag on the socket after the fetch result has been decoded. Would that resolve this issue?

still digesting this PR...

lib/protocol/protocol.js
```diff
@@ -329,18 +358,30 @@ function decodeMessageSet (topic, partition, messageSet, cb, maxTickMessages, hi
...
+              emit(null, 'message', message);
+              next(null);
+            });
           }
-          codec.decode(vars.value, function (error, inlineMessageSet) {
-            if (error) return; // Not sure where to report this
-            decodeMessageSet(topic, partition, inlineMessageSet, cb, maxTickMessages, highWaterOffset);
+          enqueue((next) => {
+            codec.decode(vars.value, function (error, inlineMessageSet) {
+              if (error) emit(error);
```
hyperlink: If there's an error from the codec we should return instead of trying to decode?

lib/kafkaClient.js
```diff
@@ -1125,6 +1125,30 @@ KafkaClient.prototype.sendFetchRequest = function (
...
+  function stateValidator (unused, type, message) {
+    const payloadMap = consumer.payloadMap;
+    if (consumer.closing || consumer.connecting || consumer.rebalancing || consumer.memberId !== memberId || consumer.generationId !== generationId) {
+      logger.error('ignoring message due to it being from an old group - memberId: ' + memberId, '!=' + consumer.memberId + ' - generationId: ' + generationId + '!=' + consumer.generationId);
+      return false;
+    }
+    if (type === 'message') {
+      const {topic, partition, offset} = message;
+      if (!payloadMap[topic] || payloadMap[topic][partition] == null) {
+        // We should have never received this in the first place
```
hyperlink: can we log an error for this case?

aikar: good idea, since old memberships should be filtered from above

lib/protocol/protocol.js
aikar: eek, yep.

aikar:
@hyperlink addressed your concerns

lib/protocol/protocol.js
```diff
@@ -226,41 +235,84 @@ function _decodeFetchResponse (resp, cb, maxTickMessages, version) {
     .word32bs('topicNum')
     .loop(decodeTopics(decodePartitions));
 
+  // Parsing is all sync, but emitting events is async due to compression
+  // At this point, the enqueue chain should be populated in order, and at
+  // the end, the 'topics' map above should be fully populated.
+  // topics is not ready for use until all events are processed, as an async
+  // decompression may add more offsets to the topic map.
+  async.series(events, (err) => {
```
hyperlink: Could we process messages in parallel instead of series? Looks like only the results (the decoded messages) need to be emitted in order.

aikar: No, that would require 2 queues of processing, and complicate the logic greatly, and not really save any meaningful amount of time.

aikar: It would require a queue per topic/partition combination, and would require waiting until every queue is done processing.

And for most cases, the processing is synchronous anyways. We only go async for compression and hitting max-messages-per-tick.

aikar:
we are running this change in production without issue, and thrown bulk load at it, including compressed messages, and it has greatly improved reliability of the client.

aikar:
@hyperlink Any hold up? This is a highly critical fix.

I'm not comfortable changing the implementation any more at this point, as this works and does not have any issues (We've been running this in production for 2 weeks now).

Any further changes should be separate PR's/issues.

thebigredgeek:
Hmm, how is that not yet merged?  Any updates here?

thebigredgeek:
@hyperlink

aikar:
@hyperlink are you going to be able to merge this soon? 

For anyone waiting on this, you should be able to `npm install https://github.com/aikar/kafka-node#fix-fetch`

thebigredgeek:
TBH I just moved to librdkafka via the nodejs bindings.  Ran into too many code smells with this lib with TS definitions and didn't have time to patch the defs unfortunately.

aikar:
@hyperlink can we get a release with this now?

Issue #1075, opened 2018-09-10
ConsumerGroup always return true on checking new partitions


makhtev:
### Environment

* Node version: v9.11.2
* Kafka-node version: 3.0.0
* Kafka version: 0.10

### For specific cases also provide
* Number of Brokers: 1
* Number partitions for topic: 44

### Include Sample Code to reproduce behavior

In `consumerGroup.js` kafka-node check for new partitions in topic:
```js
return numberOfPartitions !== _.get(metadata, `[${topic}].length`, 0);
```
If topic name in kafka is with dots (e.g.: `foo.bar`) this line of code will return `true` everytime

## Possible fix
Just add quotes around `${topic}` :
```js
return numberOfPartitions !== _.get(metadata, `['${topic}'].length`, 0);
```

PR #1076, opened 2018-09-10
Fix checking new partitions

makhtev:
Fix for issue #1075


hyperlink:
Thanks for your contribution!

PR #1081, opened 2018-09-11
add support for DescribeConfigs protocol

ahmedlafta:
This PR adds support for DescribeConfigs, which allows a user to get the config entries for either a broker or topic (aka resource).

@hyperlink let me know if you have any opinions on data format input/output.

https://kafka.apache.org/protocol#The_Messages_DescribeConfigs

### DIFF
--- a/README.md
+++ b/README.md
@@ -1148,6 +1148,83 @@ admin.createTopics(topics, (err, res) => {
 
 See [createTopics](#createtopicstopics-cb)
 
+### describeConfigs(payload, cb)
+
+Fetch the configuration for the specified resources. It requires Kafka 0.11+.
+
+* `payload`: **Array**, array of resources
+* `cb`: **Function**, the callback
+
+Example:
+
+```js
+const resource = {
+  resourceType: '2',        // '2' for topic, '4' for broker
+  resourceName: 'my-topic-name',
+  configNames: []           // specific config names, or empty array to return all,
+  includeSynonyms: false,   // requires kafka 2.0+
+}
+
+admin.describeConfigs([resource], (err, res) => {
+  console.log(JSON.stringify(res,null,1));
+})
+```
+
+Result:
+
+```json
+[
+ {
+  "configEntries": [
+   {
+    "synonyms": [],
+    "configName": "compression.type",
+    "configValue": "producer",
+    "readOnly": false,
+    "configSource": 5,
+    "isSensitive": false
+   },
+   {
+    "synonyms": [],
+    "configName": "message.format.version",
+    "configValue": "0.10.2-IV0",
+    "readOnly": false,
+    "configSource": 4,
+    "isSensitive": false
+   },
+   {
+    "synonyms": [],
+    "configName": "file.delete.delay.ms",
+    "configValue": "60000",
+    "readOnly": false,
+    "configSource": 5,
+    "isSensitive": false
+   },
+   {
+    "synonyms": [],
+    "configName": "leader.replication.throttled.replicas",
+    "configValue": "",
+    "readOnly": false,
+    "configSource": 5,
+    "isSensitive": false
+   },
+   {
+    "synonyms": [],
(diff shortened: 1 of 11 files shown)


test/test.admin.js
```diff
@@ -88,4 +89,132 @@ describe('Admin', function () {
...
   });
+
+  describe('#describeConfigs', function () {
+    const createTopic = require('../docker/createTopic');
+    let admin, consumer;
+    const topic = uuid.v4();
+    const groupId = 'test-group-id';
+
+    before(function (done) {
+      if (process.env.KAFKA_VERSION === '0.8' || process.env.KAFKA_VERSION === '0.9' || process.env.KAFKA_VERSION === '0.10') {
```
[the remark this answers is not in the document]
ahmedlafta: much cleaner 👍

test/test.admin.js
```diff
@@ -88,4 +89,132 @@ describe('Admin', function () {
...
+        this.skip();
+      }
+
+      createTopic(topic, 1, 1).then(function () {
+        consumer = new ConsumerGroup({
```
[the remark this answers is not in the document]
ahmedlafta: We don't, good catch! I've updated the test. Also, do you want me to add a `close` function onto `Admin` for convenience? It'd just delegate to the `kafkaClient`. I can do that in another PR.

README.md
```diff
@@ -1148,6 +1148,83 @@ admin.createTopics(topics, (err, res) => {
...
+Fetch the configuration for the specified resources. It requires Kafka 0.11+.
+
+* `payload`: **Array**, array of resources
+* `cb`: **Function**, the callback
+
+Example:
+
+```js
+const resource = {
+  resourceType: '2',        // '2' for topic, '4' for broker
```
aikar: should export a constant map instead, and do like resourceType: RESOURCE_TYPES.TOPIC

magic numbers as an api is bad

ahmedlafta: Make sense and i agree, i just did not see any other examples within the library already that forced library consumers to import definitions of a request variable.

"Magic numbers" are already being used in the producer:
```
// Partitioner type (default = 0, random = 1, cyclic = 2, keyed = 3, custom = 4), default 0
    partitionerType: 2
```
which I'm assuming are just the kafka defined values (which I have done the same in this PR).

As an alternative to a new constant map, we could allow users to pass in the string names of the resource type e.g. `broker`, `topic`. I have no preferences.

@aikar @hyperlink thoughts?

aikar: just because bad practice was used else where, doesn't mean it should continue :)

strings are more readable, but also vulnerable to typos.

but it could be a 
```javascript
const typeMap = {
topic: 2,
broker: 4
}

export const TYPES = {
    TOPIC: "topic",
    BROKER: "broker",
};

const type = typeMap[userOption] || userOption;

```
then the enum map can be provided, but string versions would still work (and even numbers for anyone relying on implementation details)

hyperlink: Good suggestions @aikar. 👍 

We can make it a string to number map and yield an error if it's not in the map.

ahmedlafta: Yes I love this approach, thanks @aikar ! I'll push up the changes for this soon.

ahmedlafta: Pushed up the changes for this. I added `RESOURCE_TYPES` directly onto `Admin` so user's don't have to separately import the resource types if that's the only spot that they're going to use it. Also updated the README as I had some pieces wrong, Let me know if there are any changes you'd like to see (any nitpicks as well).

I'll also probably refactor the validation logic in `kafkaClient` soon when I implement the `AlterConfigs` protocol.

jlandersen:
Accidentally did this in https://github.com/SOHU-Co/kafka-node/pull/1142 :-)

Any change this would get merged soon?

ahmedlafta:
@hyperlink any concerns? I've been using this functionality from my commit for a few months now.

hyperlink:
Thanks for your contribution @ahmedlafta !

PR #1082, opened 2018-09-11
Improvements to custom Partitioners for Producer

aikar:
Allows passing your own Partitioner object directly,
and fixes setting default to custom for high level if
you pass in a custom partitioner

Also passes the entire payload to the custom partitioner
so that you can analyze the data of the payload to make judgement


aikar:
@hyperlink any holdup?

PR #1100, opened 2018-10-15
Feature/list topics

ahmedlafta:
This seems like a basic operation that could be made easier by just providing an alias onto the Admin. Feel free to close if you don't think it's necessary.


hyperlink:
Thanks for your contribution!

Issue #1097, opened 2018-10-10
how to commit message manually?


dearpeach:
## Questions?

* Check the [Troubleshooting / FAQ](https://github.com/SOHU-Co/kafka-node#troubleshooting--faq) ?
* Search the [issues](https://github.com/SOHU-Co/kafka-node/issues) ?

### Environment

* Node version:
* Kafka-node version:
* Kafka version:

### For specific cases also provide
* Number of Brokers:
* Number partitions for topic:

### Include Sample Code to reproduce behavior

```js
// include code here
```

### Include output with [Debug](https://github.com/SOHU-Co/kafka-node#how-do-i-debug-an-issue) turned on

Thanks for your contribution!

chrisandrews7:
@dearpeach the reason you can't commit is because the [commitInterval](https://github.com/SOHU-Co/kafka-node/blob/91e361d8c012838b9931236ba0799cb04fcaca0d/lib/highLevelConsumer.js#L572) window is still enforced.
When the consumer first connects it commits, and because the `autoCommitIntervalMs` default value is `5000`, your manual commits are ignored until that time has elapsed since the previous commit.

You can use a force commit, by setting the first argument to `true`.
This will ignore the commit interval.
```js
consumer.commit(true, (err, data) => {
    console.log(err, data);
})
```

PR #1114, opened 2018-10-29
Force option on commit for highLevelConsumer in readme

chrisandrews7:
Improving docs following on from the question in https://github.com/SOHU-Co/kafka-node/issues/1097


hyperlink:
HLC has been removed but I have added it to the ConsumerGroup.

PR #1133, opened 2018-12-04
Allow specifying an encoding for ConsumerGroup (fixes #1131)

### LINKED ISSUE #1131: ConsumerGroup does not support 'buffer' encoding, corrupts binary data
### Environment

* Node version: 10.4.0
* Kafka-node version: 3.0.1
* Kafka version: 0.10.2

* Number of Brokers: 1
* Number partitions for topic: 1

```js
const { ConsumerGroup } = require('kafka-node');

var options = {
  kafkaHost: 'kafka:9092',
  groupId: 'TestGroup',
};

var consumerGroup = new ConsumerGroup(options, 'mytopic');

consumerGroup.on('message', (message) => {
  // XXX: observe that message.value is not a Buffer
  // there is no option to make it one via ConsumerGroup.
  // binary data is corrupted from the UTF8 conversion
  console.log(typeof message.value);
  console.log(message.value);
});
```

Output is that message.value is a string. If you use binary encoded data like with AVRO, then you will end up with corrupted data.

Consumer offers an encoding option, but ConsumerGroup does not. It looks like ConsumerGroup needs to include this option and allow it to be passed down.

mishan:
This allows 'buffer' to be specified so that binary data to be consumed via ConsumerGroup without corruption, resolving issue #1131

### DIFF
--- a/README.md
+++ b/README.md
@@ -722,6 +722,7 @@ var options = {
   // An array of partition assignment protocols ordered by preference.
   // 'roundrobin' or 'range' string for built ins (see below to pass in custom assignment protocol)
   protocol: ['roundrobin'],
+  encoding: 'utf8', // default is utf8, use 'buffer' for binary data
 
   // Offsets to use for new groups other options could be 'earliest' or 'none' (none will emit an error if no offsets were saved)
   // equivalent to Java client's auto.offset.reset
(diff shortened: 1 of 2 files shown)


mishan:
The CI fail seems spurious. I can't seem to retrigger a build as Travis CI doesn't give me this option.

I'm going to close + reopen this in attempt to trigger one.

PR #1163, opened 2019-01-08
remove zookeeper

hyperlink:
Drop zookeeper support from kafka-node

- [x] Update README

- [x] Remove HLC

- [x] Remove Client

- [x] Update examples

- [x] Update tests

- [x] Update typescript definitions


lib/consumerGroup.js
```diff
@@ -762,6 +733,144 @@ ConsumerGroup.prototype.removeTopics = function (topics, cb) {
...
+  }
+
+  this.needToCommit = false;
+  this.committing = true;
+  setTimeout(
+    function () {
+      this.committing = false;
+    }.bind(this),
+    this.options.autoCommitIntervalMs
+  ).unref();
```
ahmedlafta: Hey @hyperlink `unref` causes an error in an Electron environment.
```
Uncaught TypeError: setTimeout(...).unref is not a function
    at ConsumerGroup.autoCommit
``` 

Is there an alternative way to clean up the timeout reference?

hyperlink: I suppose we can check to see if `unref` is available and then call it if it's there.

PR #1201, opened 2019-02-26
Add connect event for HighLevelConsumer in types/index.d.ts

insanehong:
ConsumerGroup and ConsumerGroupStream of kafka-node had already "connect" event .
But HighLevelConsumer in types/index.d.ts missed event for HighLevelConsumer.

### DIFF
--- a/types/kafka-node-tests.ts
+++ b/types/kafka-node-tests.ts
@@ -125,6 +125,7 @@ const cgOptions: kafka.ConsumerGroupOptions = {
 
 const consumerGroup = new kafka.ConsumerGroup(cgOptions, ['topic1']);
 consumerGroup.on('error', (err) => { });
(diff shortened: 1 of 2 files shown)


types/kafka-node-tests.ts
```diff
@@ -125,6 +125,7 @@ const cgOptions: kafka.ConsumerGroupOptions = {
 
 const consumerGroup = new kafka.ConsumerGroup(cgOptions, ['topic1']);
 consumerGroup.on('error', (err) => { });
+consumerGroup.on('connect', (err) => { });
```
hyperlink: connect event does not emit with an `err`.

insanehong: Oops! resoved and send new PR

PR #1216, opened 2019-03-26
KafkaClient should not refresh metadata when broker socket is closed …

hyperlink:
…due to being idle

### DIFF
--- a/lib/kafkaClient.js
+++ b/lib/kafkaClient.js
@@ -762,6 +762,7 @@ KafkaClient.prototype.createBroker = function (host, port, longpolling) {
   });
   socket.on('close', function (hadError) {
     self.emit('close', this);
+    logger.debug(`${self.clientId} socket closed ${this.addr} (hadError: ${hadError})`);
     if (!hadError && self.closing) {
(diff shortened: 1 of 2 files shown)


crobinson42:
It appears this test fails on all builds in Travis 🔨 
#removeTopics
      1) "after all" hook: close consumer group

https://github.com/SOHU-Co/kafka-node/blob/master/test/test.consumerGroup.js#L1123

crobinson42:
This will cause:

`kafka-node:KafkaClient kafka-node-client to 192.168.0.151:9092 is idle not reconnecting +10m`

What do you think about adding a param to force reconnecting when idle in this scenario?

crobinson42:
Something like `new KafkaClient({ reconnectOnIdle: true })`

hyperlink:
> It appears this test fails on all builds in Travis 🔨
> #removeTopics
> 1) "after all" hook: close consumer group
> 
> https://github.com/SOHU-Co/kafka-node/blob/master/test/test.consumerGroup.js#L1123

Yup that is one of the flaky tests. :unamused:

> Something like `new KafkaClient({ reconnectOnIdle: true })`

Not against adding this. PR?

Issue #1019, opened 2018-07-09
Request timed out after 30000ms with increased load


skumar568:
Hi,

Our code works fine until 100 hits per 60 seconds, but fails with 200 hits per 60 seconds with below exception:-
Fri, 06 Jul 2018 10:58:41 GMT ctools     Exception occured in sendToProducer function { TimeoutError: Request timed out after 30000ms
    at new TimeoutError (/* /node_modules/kafka-node/lib/errors/TimeoutError.js:6:9)
    at Timeout.setTimeout [as _onTimeout] (/*/node_modules/kafka-node/lib/kafkaClient.js:389:14)
    at ontimeout (timers.js:482:11)
    at tryOnTimeout (timers.js:317:5)
    at Timer.listOnTimeout (timers.js:277:5) message: 'Request timed out after 30000ms' }

### Environment
* Node version: 8.10.0
* Kafka-node version: 2.4.1
* Kafka version: 

### For specific cases also provide
* Number of Brokers: 3
* Number partitions for topic: 0

### Include Sample Code to reproduce behavior
Code snippet used for creating connection and producer is attached,
[KafkaClient.txt](https://github.com/SOHU-Co/kafka-node/files/2174817/KafkaClient.txt)

Any suggestions/references/pointers are helpful to resolve this issue.

Thanks,
Sanjeev

jnotnull:
Maybe you should open the debug mode and attach the debug.log

robertjamesmiller:
we have a similar issue with publishing throughput from a single NodeJS process

JsonMa:
I also have this problem, 1000 hits per second

kabaluyot:
Any solution?

dbelieves:
Does anybody have a solution for this?

mshauny:
Has anyone being able to find a workaround this yet?

PR #1220, opened 2019-04-01
Issue/#1019 Request timed out after 30000ms with increased load

MuhammedAshrafElsayed:
This pull request should solve the issue #1019 
**The issue root cause:**
if you require a request timeout and require ack, the kafka-node will queue your callback in a queue and start a timer. This issue happens when the event queue has a lot of requests to handle -like in case of load test- so the event loop takes a lot of time to handle the recieved ack from the kafka and the timer will timeout and a timeout exception will fire although the kafka has sent the ack and the service received it successfully.
**solution**
this solution checks the return value of unqueueCallback method, if it is equal 'null' this means handleReceivedData method has handled this request and unqueue it before so it will return without firing request timeout error.

### DIFF
--- a/lib/kafkaClient.js
+++ b/lib/kafkaClient.js
@@ -487,8 +487,10 @@ KafkaClient.prototype.wrapTimeoutIfNeeded = function (socketId, correlationId, c
   }
 
   timeoutId = setTimeout(() => {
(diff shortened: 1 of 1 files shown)


kabaluyot:
this is still happening even in the latest released version of kafka-node

PR #1221, opened 2019-04-01
#1019 #comment Request timed out after 30000ms with increased load

MuhammedAshrafElsayed:
This pull request should solve the issue #1019 
**The issue root cause:**
if you require a request timeout and require ack, the kafka-node will queue your callback in a queue and start a timer. This issue happens when the event queue has a lot of requests to handle -like in case of load test- so the event loop takes a lot of time to handle the recieved ack from the kafka and the timer will timeout and a timeout exception will fire although the kafka has sent the ack and the service received it successfully.
**solution**
this solution checks the return value of unqueueCallback method, if it is equal 'null' this means handleReceivedData method has handled this request and unqueued it before so it will return without firing request timeout error.

### DIFF
--- a/lib/kafkaClient.js
+++ b/lib/kafkaClient.js
@@ -487,8 +487,10 @@ KafkaClient.prototype.wrapTimeoutIfNeeded = function (socketId, correlationId, c
   }
 
   timeoutId = setTimeout(() => {
-    let isInQueu = this.unqueueCallback(socketId, correlationId);
-    if (!isInQueu) return ;
+    let wasInQueue = this.unqueueCallback(socketId, correlationId);
+    if (!wasInQueue) {
+      return;
+    }
     callback(new TimeoutError(`Request timed out after ${timeout}ms`));
     callback = _.noop;
   }, timeout);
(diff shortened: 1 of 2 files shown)


jeremy-j-ackso:
It looks like this is only failing on Kafka 0.9, but I'd find it hard to believe it wasn't something goofy with Kafka and not the changes Can Travis be re-run for this PR so that a merge can happen? This is something I'm running in to.

lkuffo:
We are also running into this issue using kafka 2.1.1. Looking forward for this PR to be merged

rubancar:
This issue is becoming very annoying lately, when will we have this in the latest version?

hyperlink:
> Can Travis be re-run for this PR so that a merge can happen? This is something I'm running in to.

I've rerun it several times and it fails consistently.

hyperlink:
> so the event loop takes a lot of time to handle the recieved ack from the kafka and the timer will timeout and a timeout exception will fire although the kafka has sent the ack and the service received it successfully.

How do you know if the service had received it successfully? If the client called the callback for that request the timer should be cleared since unqueueCallback and [invoking the callback](https://github.com/SOHU-Co/kafka-node/blob/51b6292340b12af423c1e331ab5eda0f3beb1852/lib/baseClient.js#L584) (clearing the timer) are done synchronously.

Is there a way to create a test to reproduce this issue?

nancycg:
Any solution for this?

PR #1223, opened 2019-04-03
Made Producer & Consumer types as EventEmitter

katsanva:
This will allow using `consumer.once`, that will simplify testing.


hyperlink:
Thanks for your contribution.

Issue #1260, opened 2019-05-06
When an broker is dead, kafkaClient will constantly retrying.


qizigod:
### Environment

* Node version: 8.11.3
* Kafka-node version: 4.1.2
* Kafka version: 2.2.0

### For specific cases also provide
* Number of Brokers: 3
* Number partitions for topic: 1

### Include Sample Code to reproduce behavior
when an broker is dead, kafkaClient will constantly retrying. The recurring steps are as follows.
1. running three brokers (kafka-host:9092/ kafka-host:9093/ kafka-host:9094)
2. kafka-host:9094 is leader and others is follower
3. consumer start consume kafka-host:9094
4. shutdown kafka-host:9094
5. consumer will continue to connect kafka-host:9094 and unable to consume

kafkaClient.js -> createBroker
```js
socket.on("error", function(err) {
  this.error = err;
  if (!self.connecting) {
    self.emit("socket_error", err);
  }
});
socket.on("close", function(hadError) {
  if (!hadError && self.closing) {
    self.clearCallbackQueue(this);
  } else {
    let error = this.error;
    if (!error) {
      if (self.options.sasl && socket.saslAuthCorrelationId !== undefined) {
        delete socket.saslAuthCorrelationId;
        const message =
          "Broker closed connection during SASL auth: bad credentials?";
        error = new errors.SaslAuthenticationError(null, message);
      } else {
        error = new errors.BrokerNotAvailableError(
          "Broker not available (socket closed)"
        );
        if (!self.connecting && !brokerWrapper.isIdle()) {
          logger.debug(`${self.clientId} schedule refreshBrokerMetadata()`);
          setImmediate(function() {
            self.refreshBrokerMetadata();
          });
        }
      }
    }
    self.clearCallbackQueue(this, error);
  }
  retry(this);
});
```
when the first time kafkaClient retry connect and trying to refreshMetaData, but because of the zookeeper sessionTimeout, kafka will wating for `6000ms` instead of remove the dead broker immediately. So the kafkaClient will fetch the wrong metadata which include dead broker. And kafkaClient trying to connect a closed address which emit socket `error` event with `ECONNREFUSED` and emit socket `close` event. But `this.error` has been assigned, `refreshBrokerMetadata` will never be called, `kafkaClient` will continue retry with wrong metaData.

### Include output with [Debug](https://github.com/SOHU-Co/kafka-node#how-do-i-debug-an-issue) turned on
```shell
[...]

PR #1261, opened 2019-05-07
Bugfix/kafka client retrying

chosh31:


chosh31:
Fix: kafkaClient will constantly retrying - connectToBrokers (#1260)

PR #1274, opened 2019-05-28
add reconnect to kafka code after kafka broker down

tomgiggs:
fix:Unable to find available brokers to try
fix:Broker not available
fix:TimeoutError: Request timed out after 30000ms

I have test the code ,it can run well,if the code has error beyond my expect,please give me a simple reply


bmxpiku:
This could solve https://github.com/SOHU-Co/kafka-node/issues/1427 as well - I think...

PR #1280, opened 2019-06-06
Fix describe configs for multiple brokers (#1772)

### LINKED ISSUE #1172: DescribeConfigs for brokers not working with multiple brokers
`describeConfigs` does not work when requesting broker configs and the cluster has multiple brokers. The issue is that the implementation sends a describe config request to all brokers and merges the results. As a result requesting configs for broker X will result in errors from all other brokers than X (see [1]). 

I see a couple of things:

1) There should not be a need to send describeConfigs to all brokers. Topic requests can be handled by any broker even if that is not hosting any partitions for that topic.

2) describeConfigs for brokers should send only to the requested brokers. I recently added a sendRequestToBroker as part of #1157  which can be used to send a request to a specific broker in a safe way (handling connection setup if not ready etc.). I also started a PR for describeConfigs in #1142 where this logic was implemented - this might be inspiration for a fix. 

[1] 
 https://github.com/apache/kafka/blob/1.0/core/src/main/scala/kafka/server/AdminManager.scala#L321

### Environment

* Node version: any
* Kafka-node version: 4.0.0+
* Kafka version: 0.11.0-2.0

* Number of Brokers: > 1
* Number partitions for topic: any

```js
const kafka = require("kafka-node");

const client = new kafka.KafkaClient({
    autoConnect: true,
    connectRetryOptions: {
        retries: 1,
    },
    connectTimeout: 3000,
    kafkaHost: '' // any host in a multi-broker cluster
});

const adminClient = new kafka.Admin(client);


client.on("ready", () => {
    const resource = {
        resourceType: adminClient.RESOURCE_TYPES.broker,
        resourceName: '1', // any id in the multi-broker cluster
        configNames: []
    };

    const payload = {
        resources: [resource],
        includeSynonyms: false,
    };

    adminClient.describeConfigs(payload,
        (err, res) => {
        // Error: InvalidRequest: Unexpected broker id, expected 2, but received 1
        console.log(err);
    });
});
```

jlandersen:
This fixes describe configs when requesting broker configs for clusters with  more than 1 broker.
Instead of sending all requests to all brokers, requests for broker configs are sent to the specific broker. T
opic config requests can go to any broker (follows the logic of the Java SDK).

Also reworked a bit to fit better into api support functionality.

Fixes #1172.

### DIFF
--- a/lib/kafkaClient.js
+++ b/lib/kafkaClient.js
@@ -1379,6 +1379,12 @@ KafkaClient.prototype.describeConfigs = function (payload, callback) {
     return callback(new Error('Client is not ready (describeConfigs)'));
   }
   let err;
+
+  // Broker resource requests must go to the specific node
+  // other requests can go to any node
+  const brokerResourceRequests = [];
+  const nonBrokerResourceRequests = [];
+
   _.forEach(payload.resources, function (resource) {
     if (resourceTypeMap[resource.resourceType] === undefined) {
       err = new Error(`Unexpected resource type ${resource.resourceType} for resource ${resource.resourceName}`);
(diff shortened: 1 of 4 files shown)


hyperlink:
Thanks!

PR #1281, opened 2019-06-10
Filter out unsubscribed topic from consumerGroup.topicPartitionLength

thynson:
If two consumer group of same group id subscribed different topics,
Return value of ConsumerGroup._checkTopicPartitionChange will always
return true, resulting into leave and rejoin.

Actually I'm considered whether [code](https://github.com/SOHU-Co/kafka-node/blob/master/lib/consumerGroup.js#L244) could just be rewrite into `var topics = this.topics`, 
but it seems too risky.

P.S. The reason that two consumer group of same group id subscribed different topics 
is that we're splitting our app into microservices, and we want to be able to upgrade
or rollback seamlessly.


hyperlink:
I've done some testing and I haven't been able to recreate the leave and rejoin but it doesn't seem to be big an issue.

Doesn't this change also prevent the new topic from being assigned properly?

thynson:
> Doesn't this change also prevent the new topic from being assigned properly?

Not familiar with kafka protocol, but my goal is to prevent unsubscribed topic become key of ConsumerGroup.topicPartitionLength object, so it seems more safe to do it around [invoking of createTopicPartitionLength](https://github.com/SOHU-Co/kafka-node/blob/master/lib/consumerGroup.js#L260)?

hyperlink:
> so it seems more safe to do it around [invoking of createTopicPartitionLength](https://github.com/SOHU-Co/kafka-node/blob/master/lib/consumerGroup.js#L260)?

Yes, you could filter the topic from there and it won't prevent the assignment of those new topics. 

I am not sure of the correct behavior. Do you believe the consumer leaders own assigned topics should prevent it from assigning topics and partitions it's not aware of in the group?

thynson:
After some experiment, I think the best way to correct the check is to load all topics in `this.topicPartitionLength` instead filter out unsubscribed topic.

ismatoro:
I'm having the same problem, I want to have two different consumerGroups subscribed to two topics (one per each). Consumers start a rebalance loop for this reason, because you are requesting metadata of just the subscribed topics, and the function _checkTopicPartitionChange is comparing them with the metadata of the group, not just the consumer topics

thynson:
@hyperlink Could this PR be reviewed again if it could be merged?

PR #1327, opened 2019-09-16
Tests for Kafka version 2.12-2.1.1 and 2.12-2.2.1

hxinhan:


HansonHH:
Would be nice if kafka-node can have more tests covering Kafka upwards versions and everyone would be more confident to use kafka-node with these versions. What do you think? @hyperlink

PR #1349, opened 2019-11-04
Fix ready event emitting for producers

raymondfeng:
Make sure the ready event is emitted asynchronously so that it can be
caught by listeners.

Signed-off-by: Raymond Feng <enjoyjava@gmail.com>


raymondfeng:
@hyperlink No, I don't think so. At the moment, the `BaseProducer` constructor emit the `ready` event synchronously if the client is ready. There is no chance for other code to set up a listener as the event happens before the constructor returns.

PR #1370, opened 2020-01-03
Fixed isAsync params usage for createTopics (#900)

### LINKED ISSUE #900: The README documentation is wrong/unclear
## Bug Report
The following implies that calling `producer.createTopics(['x']. false)` blocks the execution until the Kafka's topic is created.

> When async is set to false, this method does not return until all topics are created

Which does not seem to be the case according to https://github.com/SOHU-Co/kafka-node/blob/master/lib/client.js#L369

The function uses `retry` which uses `setTimout`. It simply calls `cb(null)`, returns and keeps retrying asynchronously.

```
createTopics(topics, async, cb)
This method is used to create topics on the Kafka server. It only works when auto.create.topics.enable, on the Kafka server, is set to true. Our client simply sends a metadata request to the server which will auto create topics. When async is set to false, this method does not return until all topics are created, otherwise it returns immediately.

topics: Array, array of topics
async: Boolean, async or sync
cb: Function, the callback
Example:

var kafka = require('kafka-node'),
    Producer = kafka.Producer,
    client = new kafka.Client(),
    producer = new Producer(client);
// Create topics sync
producer.createTopics(['t','t1'], false, function (err, data) {
    console.log(data);
});
// Create topics async
producer.createTopics(['t'], true, function (err, data) {});
producer.createTopics(['t'], function (err, data) {});// Simply omit 2nd arg
```

### Environment

* Node version:
* Kafka-node version:
* Kafka version:

* Number of Brokers:
* Number partitions for topic:

```js
// include code here
```

Thanks for your contribution!

lamweili:
Fixed #900 - The README documentation is wrong/unclear

### DIFF
--- a/lib/baseClient.js
+++ b/lib/baseClient.js
@@ -253,7 +253,7 @@ Client.prototype.createTopics = function (topics, isAsync, cb) {
 
   if (typeof isAsync === 'function' && typeof cb === 'undefined') {
     cb = isAsync;
-    isAsync = true;
+    isAsync = false;
   }
 
   try {
--- a/lib/baseClient.js
+++ b/lib/baseClient.js
@@ -299,7 +299,7 @@ Client.prototype.createTopics = function (topics, isAsync, cb) {
     });
   });
 
-  if (!isAsync) {
+  if (isAsync) {
     cb(null);
   }
 };
--- a/test/test.offset.js
+++ b/test/test.offset.js
@@ -14,7 +14,7 @@ describe('Offset', function () {
     client = new Client();
     producer = new Producer(client);
     producer.on('ready', function () {
-      producer.createTopics(['_exist_topic_3_test'], true, function (err) {
+      producer.createTopics(['_exist_topic_3_test'], false, function (err) {
         done(err);
       });
     });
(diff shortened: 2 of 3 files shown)


peteriman:
Fixed: The README documentation is wrong/unclear (#900)

lamweili:
@hyperlink 

I noticed that most of the pull requests have the following error from the Travis-CI test report:

>   1) SSL Producer #send "before all" hook:
>      Uncaught AssertionError: expected { '0': 14 } to have property '0' of 0 (got 14)
>       at Assertion.fail (node_modules/should/lib/assertion.js:183:17)
> 

Some examples are #1612.4 and #1612.7 of
https://travis-ci.org/SOHU-Co/kafka-node/builds/632764529?utm_medium=notification&utm_source=github_status

This error has got nothing to do with the pull requests, but rather a test suite issue.

I have patched test.producer.js to use uuid.v4() instead of Date.now() in the topic name, to reduce the risk of collision (8526b0bede5aeb9508fb8c4a3f05ce6f43bdfd94). Especially when Travis-CI runs multiple environments concurrently, it is highly likely that Date.now() might not be unique enough.

Issue #1119, opened 2018-11-06
no Admin class in typescript index.d.ts


huanhuang-wang:
how can I use Admin in typescript

opsb:
As a workaround you can do
```
const kafka = require("kafka-node");
const KafkaAdmin = kafka.Admin;
```

vakho10:
> 
> 
> As a workaround you can do
> 
> ```
> const KafkaAdmin = require("kafka-node").Admin;
> ```

Hi, could you guys add the definition? This doesn't work for me in Angular project for instance :(

PR #1440, opened 2020-12-27
Fix add types for Admin and fix some type error

Tseian:
issue: #1119


ramakrishna-battala-ck:
hello @hyperlink can we get this PR in. This is a feature we are also trying to make use of.

